PR #3, opened 2024-07-09
🧪 Adds Github CI workflow

hardbyte:
Adds a very basic CI using Github workflows to ensure PRs don't break the build or break tests.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  rust:
+    name: Rust format, lint, and test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      # https://github.com/actions-rs/toolchain
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
avkonst, line 20: the rust version is specified in the file in the repo root.. I assume that should be used here?

hardbyte: No because this is just the version of rustup toolchain (stable, nightly or a specific version).

hardbyte: By default it will read the version from the toolchain file but we just need to rename it (pushing now).

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

avkonst: ahh it installs it on fresh ubuntu... would be interesting to see if it has any chaching...

hardbyte:
@avkonst I assume that `cargo test` is failing for you and is not just a Github CI specific thing?

andyyu2004:
@hardbyte I hope you don't mind me rebasing this branch. Tests are still failing due to formatting. I've had a quick discussion with @avkonst about formatting the output code in a way that avoids updating mtimes unnecessarily. I'll implement that and then this branch should be mergeable!

PR #21, opened 2024-08-11
HTTP execution tests

andyyu2004:
This takes the existing `reflectapi-demo/*.http` examples and turns them into tests by executing them and taking a snapshot.

Majority of the diff is splitting the crate into bin & lib. See the 2nd commit for relevant stuff.

With further `.http` files this will cover gaps (i.e. runtime behavior) that the existing tests won't cover e.g. #17 had no associated tests.


avkonst:
very nice, thanks.. maybe more *.http files to include for completeness?

andyyu2004:
Yes, definitely. I think this PR is just to introduce the infrastructure, any future code changes should add more test files. I also wanted this to help test the coming msgpack implementation.

PR #27, opened 2024-08-13
Typecheck generated code

andyyu2004:
#15 added a formatting which provided syntactic checks and uncovered a few issues. This PR type-checks the generated code too which is useful for catching more issues when working on the code generator.

Only 1 existing issue was caught!

Side effect of these two PRs is that running tests is a bit slow as both `prettier` and `tsc` have terrible startup times. We can move to a [faster formatter](https://github.com/biomejs/biome) but there isn't really an alternative to `tsc` that I'm aware of.

cc @JosiahBull Franklin tells me you have experience with custom rust builds like this. I wonder if you have better ideas than this implementation?


andyyu2004:
Just going to merge this to avoid blocking some of the other PRs. This can be revisited anytime.

PR #30, opened 2024-08-18
OpenAPI spec generation

### LINKED ISSUE #22: codegen for openapi language
We need a new "language" codegen enabled: output to openapi.json - the schema file which is used for documenting APIs by the tools like https://redocly.github.io/redoc/ 
We are not planning to do anything else with openapi schema, only for documentation purposes.

andyyu2004:
This implements a `reflectapi.json` to `openapi.json` conversion.

I have renamed type `parameters` to type `arguments` mostly to make me less confused. I do think it's the more correct name but can revert that change if wanted.

I've also renamed `()` to `unit` to avoid non alphanumeric type names which OpenAPI won't allow. 

This is still in draft phase and there a bunch of hacks that need to be solved.

Resolves #22


reflectapi-demo/librust_out.rmeta
avkonst, line 1: what is this file?

reflectapi/src/codegen/rust.rs
avkonst, line 1: Similarly to rust and typescript, reflectapi-cli should allow to generate openapi.json from the reflectapi.json. Relates to my comment about moving the implementation to codegen.

reflectapi-schema/src/lib.rs
```diff
@@ -18,6 +21,8 @@ pub struct Schema {
     pub output_types: Typespace,
 }
 
+type Subst = HashMap<String, TypeReference>;
```
avkonst, line 24: Is it used in many places? can we use the HashMap<> directly there... it was very difficult to read the code where it is used.

reflectapi-schema/src/lib.rs
```diff
@@ -473,10 +478,30 @@ impl TypeReference {
 
     fn rename_type(&mut self, search_string: &str, replacer: &str) {
         self.name = rename_type_or_module(&self.name, search_string, replacer);
-        for param in self.parameters.iter_mut() {
+        for param in self.arguments.iter_mut() {
             param.rename_type(search_string, replacer);
         }
     }
+
+    fn instantiate(self, subst: &Subst) -> TypeReference {
```
avkonst, line 486: where is it used? what are you trying to do here? This seems like belongs to codegen... It is also very confusing that TypeReference can be not instanciated. It is always instanciated, TypeParameter is not....

avkonst (changes requested):

reflectapi-demo/librust_out.rmeta
andyyu2004: damn, must have accidentally left it behind as a part of #27

reflectapi-schema/src/lib.rs
andyyu2004: It's used in code like the following. There can implicitly be generics in `TypeReference`s which this function tries to replace with types. `TypeReference::instantiate` is called indirectly by `Struct::instantiate` and `Enum::instantiate` which call `Field::instantiate`. 

I do think that the representation of types is not ideal, in particular there is no way to distinguish between a type named `T` and `T` representing a reference to a type parameter. This was the most minimal code I could come up with that mostly works.

```rust
            crate::Type::Struct(strukt) => match self.struct_to_schema(
                schema,
                kind,
                &strukt.clone().instantiate(&ty_ref.arguments),
            )
```

avkonst: and where the Struct::instantiate is used?
in general you can not implement this sort of functionality without having access to the type definition.

avkonst:
"I've also renamed '() to unit to avoid non alphanumeric type names which OpenAPI won't allow." - that really should be a concern of OpenAPI codegen block. It needs to do name remapping and generic resolutions on its own without involving the changes to the schema.

reflectapi-schema/src/lib.rs
avkonst: I think I understand what you are doing here. I think I do something similar in the TS and Rust codegen. Check resolve type ref functions.

andyyu2004: The above code sample is verbatim from `codegen/openapi.rs`. Given a type reference with args `S<int, bool>` referencing generic struct `S<T, U> { t: T, u: U }`. Instantiate takes a generic struct with `k` type parameters and `k` type arguments and returns a non-generic struct with the parameters filled in.

`instantiate(S<T, U> { t: T, u: U}, [int, bool]) returns non generic struct S { t: int, u: bool }`.

andyyu2004: Without the instantiation, the openapi spec will incorrect state that `S.t` has type `T` instead of `int` or whatever.

andyyu2004:
Generally agree with the renaming point. However in this case, I'm not convinced `()` is a good name considering all other types have alphanumeric friendly names including tuples (TupleN). We could even call it `Tuple0` for consistency?

avkonst:
> Generally agree with the renaming point. However in this case, I'm not convinced `()` is a good name considering all other types have alphanumeric friendly names including tuples (TupleN). We could even call it `Tuple0` for consistency?

Ok, agree with either Unit or Tuple0

reflectapi-schema/src/lib.rs
avkonst: I think this code belongs to codegen... Maybe some helper wrap struct is needed, not sure. It feels very foreign to have it in the schema.

andyyu2004: Ok agree that I didn't like `instantiate` being a public function. I'll move the instantiation functions as private free functions.

reflectapi/src/codegen/openapi.rs
```diff
@@ -0,0 +1,595 @@
...
+                            items: Box::new(self.convert_type_ref(
+                                schema,
+                                kind,
+                                ty_ref.arguments().next().unwrap(),
+                            )),
+                            unique_items: prim.name() == "std::collections::HashSet",
+                        }
+                    }
+                    // Treat as transparent wrappers? Not sure why these types are relevant to an API anyway @Andrey?
+                    "std::boxed::Box" | "std::cell::Cell" | "std::cell::RefCell"
```
andyyu2004, line 373: Can you have a look at this @avkonst? My opinion is that these pointer/wrapper types `Wrapper<T>` should be represented in the API the same as `T`. It seems irrelevant to the API how they are represented in memory.

avkonst: for openapi and typescript codegen, use fallback type whenever it is available on the type definition. These pointers fallback to T. Have a look how TS codegen resolves these. Same applies to openapi.
The reason we have to keep these explicitly defined (unless a user tells to transform to fallback explictly with an attribute), is because Rust recursive references would not work for the rust codegen. For example Struct of T with field of type T is not possible.

andyyu2004:
I think this is good enough to merge. For sure the code generation can be improved but the public API will remain stable.

reflectapi-schema/src/lib.rs
```diff
@@ -938,6 +938,14 @@ impl Field {
         self.name.as_str()
     }
 
+    pub fn is_named(&self) -> bool {
+        !self.is_unnamed()
+    }
+
+    pub fn is_unnamed(&self) -> bool {
+        self.name.parse::<u64>().is_ok()
```
avkonst, line 946: you are certain we do not miss serde_name usage here? just curious why these were needed for names but not for serde names

reflectapi/src/codegen/mod.rs
```diff
@@ -25,3 +29,99 @@ fn tmp_path(src: &str) -> PathBuf {
...
+    fn instantiate(self, args: &[TypeReference]) -> Self;
+}
+
+impl Subst for TypeReference {
+    fn subst(self, subst: &HashMap<String, TypeReference>) -> Self {
+        match subst.get(&self.name) {
+            Some(ty) => {
+                assert!(
+                    self.arguments.is_empty(),
+                    "type parameter cannot have type arguments (no HKTs)"
```
avkonst, line 47: better not to use acronyms. Use full name

reflectapi/src/codegen/mod.rs
```diff
@@ -25,3 +29,99 @@ fn tmp_path(src: &str) -> PathBuf {
 
     std::env::temp_dir().join(format!("reflectapi-{hash}"))
 }
+
+trait Subst {
```
avkonst, line 33: Is there better more meaningful name? Looking at it I do not understand what it does

andyyu2004: How about `Substitute` in full + some comments?

reflectapi-schema/src/lib.rs
andyyu2004: Didn't think too hard about it. This is copied from `rust/ts` codegen. I think the clearest solution would be to make `name: Option<String>` rather than checking if it's a number or not.

reflectapi/src/codegen/mod.rs
andyyu2004: Just removed the higher-kinded types reference, not particular useful.

reflectapi-schema/src/lib.rs
avkonst: I see what you mean here. It is fine that names are numbers. It is by design. It captures full intent. Codegen can interpret it as it needs to.

reflectapi/src/codegen/mod.rs
avkonst: Maybe. I still not quite understand what it does. Perhaps some examples would help?

andyyu2004: I've tried to add more comments, not finding it particularly easy to document `Substitute`. I've added more comments for `Instantiate` which provides the context for why substitution makes sense. Let me know if that makes sense.

andyyu2004: @avkonst are you sufficiently happy with this to merge?

reflectapi-schema/src/spec.rs
avkonst, line 1: I think this file can be removed?

avkonst (approved):

PR #32, opened 2024-08-19
Avoid defining reflectapi::Result

andyyu2004:
I think the better workaround is to define our own trait instead. I don't think this change should break any users as `reflectapi::Result` was mostly used internally.

### DIFF
--- a/reflectapi/src/builder/handler.rs
+++ b/reflectapi/src/builder/handler.rs
@@ -85,7 +85,7 @@ where
     where
         F: Fn(S, I, H) -> Fut + Send + Sync + Copy + 'static,
         Fut: std::future::Future<Output = R> + Send + 'static,
-        R: Into<crate::Result<O, E>> + 'static,
+        R: crate::IntoResult<O, E> + 'static,
         I: crate::Input + serde::de::DeserializeOwned + Send + 'static,
         H: crate::Input + serde::de::DeserializeOwned + Send + 'static,
         O: crate::Output + serde::ser::Serialize + Send + 'static,
@@ -164,7 +164,7 @@ where
         E: crate::Output + serde::ser::Serialize + crate::StatusCode,
         F: Fn(S, I, H) -> Fut,
         Fut: std::future::Future<Output = R> + Send + 'static,
-        R: Into<crate::Result<O, E>>,
+        R: crate::IntoResult<O, E>,
     {
         let mut input_headers = input.headers;
 
@@ -253,7 +253,7 @@ where
         };
 
         let output = handler(state, input, input_headers).await;
-        let output: crate::Result<O, E> = output.into();
+        let output = UntaggedResult::from(output.into_result());
 
         let output_serialized = match content_type {
             ContentType::Json => serde_json::to_vec_pretty(&output).map_err(|err| err.to_string()),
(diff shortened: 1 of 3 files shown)


andyyu2004:
This actually causes another issue where `std::result::Result<T, E>` doesn't have an `untagged` repr. Not really sure how to workaround that without a separate type.

avkonst:
> This actually causes another issue where `std::result::Result<T, E>` doesn't have an `untagged` repr. Not really sure how to workaround that without a separate type.

Yes the response serialization layer needs to know when it serializes Ok and when Err case.

avkonst:
Perhaps it is possible to incorporate serialization of output into serialization of Ok case and Err case? i.e. move serialization to each of the case blocks here? 
image

andyyu2004:
Your comment immediately gave me the obvious idea which is to convert types just before serialization to an untagged result. I think this is a reasonably nice solution.

avkonst:
Either this, or more performant variant with reusing the existing match block just after the current serialize call.

avkonst:
BTW, I just realized that MsgPack support you added was not quite complete. I do not remember if the schema declares supported types? And the client (at least in Rust) would need to request msgpack instead.

reflectapi/src/builder/result.rs
```diff
@@ -2,60 +2,18 @@ pub trait StatusCode {
...
-#[derive(Debug, serde::Serialize)]
-#[serde(untagged)]
-pub enum Result<T, E>
-where
-    T: crate::Output,
-    E: crate::Output + StatusCode,
-{
-    Ok(T),
-    Err(E),
+pub trait IntoResponse<O, E> {
```
avkonst, line 5: I think it should be renamed to IntoResult

reflectapi/src/builder/handler.rs
```diff
@@ -278,10 +278,43 @@ where
...
+            code,
             body: bytes::Bytes::from(output_serialized),
             headers: response_headers,
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize)]
+#[serde(untagged)]
+enum UntaggedResult<T, E> {
```
avkonst, line 308: Looking at this again, I really thing we do not need this extra type. It can be simply embedded into here (which is only 1 line down)
image

andyyu2004: If I'm not mistaken, I would have to copy paste the serialization logic into the `Ok` and `Err`branches? Could you give a code sample not sure I'm following.

avkonst: Yes, repeat to_vec_pretty call in both branches of the match statement. One will need to do it for Ok value, another for error value.

andyyu2004: I don't understand how that is a more appealing solution. This is efficient (compiler probably can just transmute if it's smart enough) and fairly clean.

avkonst: This is a matter of preference, not going to argue here. You can leave it as is if it is clearer. In my opinion explicit serialization for either result or error would be clearer to read for new developer unfamiliar with untagged story. As it that solution was not immediately obvious to you, it would not be obvious immediately clear to new reader. Although new wrapping struct approach is maybe more rusty way, which is also questionable.

andyyu2004: I'm objecting mainly because it's not just a few lines of duplication. It's the msgpack/json branch and the serialization failure handling branch too. I think I will leave as the extra type for now as you're willing to accept it.

avkonst: ok.. msgpack is not a strong reason, but I am ok with that, we need to move on with it :)

andyyu2004: The main thing I didn't want to copy is the following. Anyway, I will commit to the decision for now. Can revisit in the future.

```rust
        let output_serialized = match output_serialized {
            Ok(r) => r,
            Err(err) => {
                response_headers
                    .insert("content-type", http::HeaderValue::from_static("text/plain"));
                return HandlerOutput {
                    code: http::StatusCode::INTERNAL_SERVER_ERROR,
                    body: bytes::Bytes::from(
                        format!("Failed to serialize response body: {err}").into_bytes(),
                    ),
                    headers: response_headers,
                };
            }
        };
```

PR #35, opened 2024-08-28
fix(codegen openapi): treat enum variants with a single field as transparent

andyyu2004:
They would be represented as a single item tuple which is incorrect.


avkonst:
approved

PR #38, opened 2024-09-02
Move client error type from generic to associated type

andyyu2004:
`Error` as an associated type is the correct place to put this parameter. A type will never need to implement the `Client` trait over more than one error type. 

This cleans up the code a decent amount (less generics and phantom data) and will reduce any potential type inference issues.


avkonst:
nice one. Approved

Issue #1, opened 2024-05-13
Add validation rule to discard schemas which can produce non serializable structs


avkonst:
![image](https://github.com/thepartly/reflectapi/assets/3971413/b5cc64c4-bf80-48e3-a4a3-c68063a291ea)

+ see TODO in code near the TestEnumTag test

PR #40, opened 2024-09-05
Codegen enum representation fixes

andyyu2004:
- **codegen(openapi): fix for externally tagged enums with no data**
- **preserve in `reflectapi.json` the type of fields: unnamed|named|none**


reflectapi-demo/clients/rust/generated/src/generated.rs
```diff
@@ -1,3 +1,4 @@
+/* <----- */
```
avkonst, line 1: is it possible to move this line 3 lines down?

andyyu2004: good point, I'll just push it after the header which has a few extra lines but no big deal.

reflectapi-demo/src/tests/serde.rs
```diff
@@ -110,7 +110,8 @@ fn test_enum_untagged() {
 // see https://github.com/thepartly/reflectapi/issues/1
 enum TestEnumTag {
     Variant1 { field_name: u8 },
-    Variant2(u8),
+    Variant2 { field_name: u8 },
+    // Variant2(u8),
```
avkonst, line 114: why do we need to change test code?

andyyu2004: It hits an assertion I put in. There's no point allowing this since serde-json will panic if we try to use it. https://github.com/thepartly/reflectapi/issues/1

andyyu2004: I can relax the assertion if you prefer.
```
                match &mut strukt.fields {
                    reflectapi_schema::Fields::Named(fields) => fields.push(tag_field),
                    reflectapi_schema::Fields::None => {
                        strukt.fields = reflectapi_schema::Fields::Named(vec![tag_field])
                    }
                    reflectapi_schema::Fields::Unnamed(_) => {
                        panic!("internal repr with unnamed fields is not allowed")
                    }
                }
```

avkonst: Ahh... are you planning to address it later? I think this code sample needs to move to the test suite where we test compilation error messages.

andyyu2004: I can try along with the other compile error. I'll discuss offline. The problem is I imagine serde would have implemented the compile time error already if it were easy.

avkonst: I think serde just missed it. It is not difficult to implement.

avkonst: ok, merge this and address denying at compile time in the separate MR

PR #44, opened 2024-09-11
codegen(rust): impl fmt::Debug for generated error type

andyyu2004:


reflectapi/src/codegen/rust-dependency-stubs/http.rs
avkonst, line 1: what are these stubs for?

avkonst (approved):

reflectapi/src/codegen/rust-dependency-stubs/http.rs
andyyu2004: This is how rust is typechecked. It is a big hack but it was really useful to debug generated code.  See `fn typecheck(src: &str) -> anyhow::Result<()>` in `codegen/rust.rs`.

This ensures that all the test cases generate code that actually compiles.

PR #45, opened 2024-09-11
move rust helper code to `reflectapi::rt`

andyyu2004:
Instead of inlining all helper code into the generated file. Move to a separate runtime dependency crate named `reflectapi-rt`.

This has a few benefits:
- Much, much easier to modify runtime code as it's just a usual rust file rather than within some string.
-  Easier for generated clients to use as it groups all the runtime dependencies into one crate.

PR #46, opened 2024-09-15
Use url type for url things

andyyu2004:
On top of #45 .

### DIFF
--- a/reflectapi-demo/clients/rust/generated/src/generated.rs
+++ b/reflectapi-demo/clients/rust/generated/src/generated.rs
@@ -17,29 +17,43 @@ pub mod interface {
         pub health: HealthInterface<C>,
         pub pets: PetsInterface<C>,
         client: C,
-        base_url: std::string::String,
+        base_url: reflectapi::rt::Url,
     }
 
     impl<C: reflectapi::rt::Client + Clone> Interface<C> {
-        pub fn new(client: C, base_url: std::string::String) -> Self {
-            Self {
-                health: HealthInterface::new(client.clone(), base_url.clone()),
-                pets: PetsInterface::new(client.clone(), base_url.clone()),
+        pub fn try_new(
+            client: C,
+            base_url: reflectapi::rt::Url,
+        ) -> std::result::Result<Self, reflectapi::rt::UrlParseError> {
+            if base_url.cannot_be_a_base() {
+                return Err(reflectapi::rt::UrlParseError::RelativeUrlWithCannotBeABaseBase);
+            }
+
+            Ok(Self {
+                health: HealthInterface::try_new(client.clone(), base_url.clone())?,
+                pets: PetsInterface::try_new(client.clone(), base_url.clone())?,
                 client,
                 base_url,
-            }
+            })
         }
     }
 
     #[derive(Debug)]
     pub struct HealthInterface<C: reflectapi::rt::Client + Clone> {
         client: C,
-        base_url: std::string::String,
+        base_url: reflectapi::rt::Url,
     }
 
     impl<C: reflectapi::rt::Client + Clone> HealthInterface<C> {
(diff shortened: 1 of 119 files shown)


reflectapi-demo/clients/rust/src/main.rs
```diff
@@ -22,6 +25,9 @@ async fn main() {
             Error::Network(e) => {
                 println!("Network error: {:?}", e)
             }
+            Error::UrlParse(e) => {
```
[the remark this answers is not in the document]
andyyu2004: I would need to bound `NE` by something then. I was thinking maybe we should just `unwrap` since this should never fail.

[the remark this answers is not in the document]
andyyu2004: I've changed to `expect`, it shouldn't happen unless generated code is broken.

reflectapi-demo/clients/rust/generated/src/generated.rs
```diff
@@ -49,8 +49,9 @@ pub mod interface {
         ) -> Result<reflectapi::Empty, reflectapi::rt::Error<reflectapi::Empty, C::Error>> {
             reflectapi::rt::__request_impl(
                 &self.client,
-                &self.base_url,
-                "/health.check",
+                self.base_url
```
avkonst, line 52: van base_url cause the expect to fail?

andyyu2004: It can. I can add an `assert!(!base_url.cannot_be_a_base())` in `Interface::new()`. That should catch most causes of panics early. I still it's fine to panic since it's a user error to pass a bad url that would cause `join` to fail.

```
        if let Some(base_url) = self.base_url {
            if input.starts_with('#') {
                self.fragment_only(base_url, input)
            } else if base_url.cannot_be_a_base() {
                Err(ParseError::RelativeUrlWithCannotBeABaseBase)
            } else {
                let scheme_type = SchemeType::from(base_url.scheme());
                if scheme_type.is_file() {
                    self.parse_file(input, scheme_type, Some(base_url))
                } else {
                    self.parse_relative(input, scheme_type, base_url)
                }
            }
        } else {
            Err(ParseError::RelativeUrlWithoutBase)
        }

```

avkonst: I do not like panicking or asserting in general, but especially for user input "it's fine to panic since it's a user error to pass a bad url"... 

Previously we submitted that to the networking layer (reqwest) and it handled invalid input responding with can not connect or something. Not sure why we needed this change... If we keep it the only way to handle it is to validate the base url at the place where it is supplied, meaning Interface::new() would need to become Interface::try_new() ?

avkonst: having url used instead of string would still have that problem as trailing slashes and double slashes are allowed in url.

andyyu2004: yes, but `"localhost/".join("/api")` is handled correctly. `format("{}{}", "localhost/", "/api")` results in double slash.

andyyu2004: More generally I think it's just using the right type for the job. I usually avoid string for everything except things where it's actually an arbitrary string such as a `description` because there are basically no restrictions.

avkonst: "I would much rather code panic early on error than fail in strange ways like this." - ok, then maybe the best is to assert and return an error when the base url is provided, i.e. from Interface::try_new perhaps... and assert in the handlers, but that failing assert would fire because of our bug not because of user's input

andyyu2004: changed to `try_new` and will panic in handlers.

PR #47, opened 2024-09-15
impl `Error` for `.build()` return 

### LINKED ISSUE #34: Errors returned should implement `std::error::Error`
For example, the error type in `Result<(crate::Schema, Vec<Router<S>>), Vec<crate::ValidationError>>` does not implement `Error` which makes it requires manual handling.

andyyu2004:
Resolves #34

### DIFF
--- a/reflectapi-demo/Cargo.toml
+++ b/reflectapi-demo/Cargo.toml
@@ -28,12 +28,12 @@ serde = { version = "1.0.197", features = ["derive", "rc"] }
 serde_json = "1.0.114"
 
 [dev-dependencies]
+anyhow = "1"
 trybuild = "1.0.0"
 insta = { version = "1.36.1", features = ["json"] }
 tower = "0.4.0"
 http-rest-file = "0.5.1"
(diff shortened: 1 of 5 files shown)


reflectapi-demo/Cargo.toml
```diff
@@ -26,14 +26,14 @@ chrono = { version = "0.4.37", features = ["serde"] }
 bytes = "1.5.0"
 serde = { version = "1.0.197", features = ["derive", "rc"] }
 serde_json = "1.0.114"
+anyhow = "1.0.81"
```
avkonst, line 29: is it possible to keep the demo without this dependency? no particularly strong concern, but I tried to keep the demo as much as possible with little dependencies for others to understand it quicker even if they are not familiar with RUst ecosystem libs very well

andyyu2004: Switched to `Box<dyn Error>`. We should probably also consider moving most of the tests out of `demo` into `reflectapi`.

avkonst:
approved

PR #50, opened 2024-09-17
return error on rename_types failure instead of panicking

andyyu2004:


reflectapi/src/builder/mod.rs
```diff
@@ -16,6 +17,8 @@ where
     handlers: Vec<crate::Handler<S>>,
     merged_handlers: Vec<(String, Vec<crate::Handler<S>>)>,
     validators: Vec<fn(&crate::Schema) -> Vec<crate::ValidationError>>,
+    allow_redundant_renames: bool,
+    errors: Vec<Box<dyn Error + Send + Sync>>,
```
avkonst, line 21: can these errors reuse the same types / way of doing things as validators?

reflectapi/src/validation.rs
```diff
@@ -1,28 +1,5 @@
...
-            writeln!(f, "{err}")?;
-        }
-        Ok(())
-    }
-}
-
-impl std::error::Error for ValidationErrors {}
-
 #[derive(Debug)]
 pub struct ValidationError {
```
avkonst, line 4: can this be referenced and used in the BuildErrors?

andyyu2004: Do you basically mean replace `Box<dyn Error>` with `enum BuildError {Validation(..), ... }`?

andyyu2004: Pushed a commit that moves in this direction

reflectapi/src/builder/mod.rs
```diff
@@ -245,3 +267,47 @@ impl RouteBuilder {
         self
     }
 }
+
+#[derive(Debug)]
+pub enum BuildError {
+    Validation(crate::ValidationError),
+    Other(Box<dyn Error + Send + Sync>),
```
avkonst, line 274: have not found where this variant is used

avkonst:
left question, happy overall

reflectapi/src/builder/mod.rs
andyyu2004:
```
        for validator in self.validators.iter() {
            for err in validator(&self.schema) {
                self.errors.push(BuildError::Validation(err));
            }
        }

        match glob.parse::<reflectapi_schema::Glob>() {
            Ok(pattern) => self.rename_types(&pattern, replacer),
            Err(err) => {
                self.errors.push(BuildError::Other(
                    format!("invalid glob pattern: {err}").into(),
                ));
                self
            }
        }


        if self.schema.rename_types(pattern, to) == 0 && !self.allow_redundant_renames {
            self.errors.push(BuildError::RedundantRename {
                pattern: pattern.to_string(),
            });
        }

```

PR #51, opened 2024-09-19
codegen(openapi): miscodegen of externally tagged newtype enum as untagged

andyyu2004:


avkonst:
not looking much. approved

PR #55, opened 2024-09-24
codegen(openapi): respect struct flattening

andyyu2004:


avkonst:
did not look closely, approved

PR #57, opened 2024-09-24
consolidate and fold schema in build

andyyu2004:
I've changed `build` always apply `fold_transparent_types` and `consolidate_types` , I don't any reason not to?

This is useful to write a reproducing test case for #56.


avkonst:
Shall it be builder config? "I've changed build always apply fold_transparent_types and consolidate_types , I don't any reason not to?"

andyyu2004:
I'm not sure, these feel like almost mandatory transformations. Is there any time you wouldn't want this?

avkonst:
fold_transparent_types - not sure about it.... It might be wanted for generated client code readability when aliasing types. ?

andyyu2004:
I don't mind reverting this to make it opt-in. I just need it turned on for tests.

avkonst:
I am not sure either. You can keep it as hard code behavior for now, and maybe we will be able to add it as configurable option. Alternatively it might be we could add special reflectapi attribute to force keeping wrappers in the future.

PR #58, opened 2024-10-07
feat(codegen rust): additional derives

### LINKED ISSUE #43: Explore how to add custom derive attributes to the generated client rust type
image

andyyu2004:
Resolves #43


reflectapi-derive/src/tokenizable_schema.rs
```diff
@@ -238,6 +240,35 @@ impl<'a> ToTokens for TokenizableEnum<'a> {
                 parameters: vec![#(#parameters),*],
                 representation: #representation,
                 variants: vec![#(#variants),*],
+                codegen_config: #codegen_config,
+            }
+        });
+    }
+}
+
+pub(crate) struct TokenizableLanguageSpecificTypeCodegenConfig<'a>(
```
andyyu2004, line 249: A proc macro noob at the moment. But having to manually transforming everything to tokens manually is fairly brutal? I'm hoping there is an easier way

reflectapi/src/codegen/openapi.rs
```diff
@@ -466,28 +466,19 @@ impl Converter {
...
-                    }
-                    // Treat these as transparent wrappers.
-                    "std::boxed::Box" | "std::cell::Cell" | "std::cell::RefCell"
-                    | "std::sync::Mutex" | "std::sync::RwLock" | "std::rc::Rc"
-                    | "std::rc::Weak" | "std::sync::Arc" | "std::sync::Weak" => {
-                        return self.convert_type_ref(
+                    "std::vec::Vec"
+                    | "std::array::Array"
+                    | "std::collections::HashSet"
+                    | "std::collections::BTreeSet" => Type::Array {
```
avkonst, line 472: I see 2 issues here:
1. as far as I remember HashSet has fallback type defined (which should be Vec), and so the codegen should use fallback type instead of the original type. Otherwise, codegen code would need to be changed every time when a new rust specific data structure is added. Codegen should allow to work with new types which it does not know about but have fallbacks.
2. BTreeSet does not have the fallback type defined?

reflectapi/src/codegen/openapi.rs
```diff
@@ -466,28 +466,19 @@ impl Converter {
...
                             kind,
                             ty_ref.arguments().next().unwrap(),
-                        )
-                    }
+                        )),
+                        unique_items: prim.name().ends_with("Set"),
+                    },
                     // There is no way to specify the key type in OpenAPI, so we assume it's always a string (unenforced).
-                    "std::collections::HashMap" => Type::Map {
+                    "std::collections::HashMap" | "std::collections::BTreeMap" => Type::Map {
```
avkonst, line 481: same here

reflectapi/src/codegen/openapi.rs
```diff
@@ -466,28 +466,19 @@ impl Converter {
...
-                                kind,
-                                ty_ref.arguments().next().unwrap(),
-                            )),
-                            unique_items: prim.name() == "std::collections::HashSet",
-                        }
-                    }
-                    // Treat these as transparent wrappers.
-                    "std::boxed::Box" | "std::cell::Cell" | "std::cell::RefCell"
-                    | "std::sync::Mutex" | "std::sync::RwLock" | "std::rc::Rc"
-                    | "std::rc::Weak" | "std::sync::Arc" | "std::sync::Weak" => {
```
avkonst, line 482: why did this disappear?

reflectapi/src/codegen/typescript.rs
```diff
@@ -1102,6 +1102,7 @@ fn build_implemented_types() -> HashMap<String, String> {
 
     implemented_types.insert("std::vec::Vec".into(), "Array<T>".into());
     implemented_types.insert("std::collections::HashMap".into(), "Record<K, V>".into());
+    implemented_types.insert("std::collections::BTreeMap".into(), "Record<K, V>".into());
```
avkonst, line 1105: this type is lacking fallback... adding BTreeMap support should not affect typescript codegen and any other codegen code in general.

reflectapi/src/codegen/openapi.rs
andyyu2004: Ok, I will change it such that `BTree` family falls back to the `Hash`es. The reason `Hashset` doesn't fallback is because we can set the `unique` property on openapi arrays

reflectapi/src/codegen/openapi.rs
andyyu2004: Don't need because of fallbacks

PR #60, opened 2024-10-08
Option to hide routes from documentation

andyyu2004:
Note for future, include and exclude tags also makes sense for generated clients but not implemented here.

### DIFF
--- a/reflectapi-demo/reflectapi.json
+++ b/reflectapi-demo/reflectapi.json
@@ -9,7 +9,10 @@
       "serialization": [
         "json"
       ],
-      "readonly": true
+      "readonly": true,
+      "tags": [
+        "internal"
+      ]
     },
     {
       "name": "pets.list",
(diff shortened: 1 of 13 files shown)


reflectapi-demo/reflectapi.json
```diff
@@ -9,7 +9,8 @@
       "serialization": [
         "json"
       ],
-      "readonly": true
+      "readonly": true,
```
avkonst, line 12: can we generalize this and introduce tags field as a vector of strings? Then we would be able to include / exclude routes / types and fields by tags by providing --include-tags and --exclude-tags arguments to codegen tools.

andyyu2004: Mostly done, do you expect this to apply to rust/ts too?

avkonst: possibly eventually, most likely TS, C# and PHP. I can imagine we can offer generated SDKs to customers by including only features which are tagged for that customer (or group of customers). We have got the plans to narrow down the scope of APIs available significantly

avkonst: but not high priority at the moment

PR #62, opened 2024-10-25
feat(json): add impl traits for serde_json::Map type.

JosiahBull:


andyyu2004 (commented):
Just leave the version alone,  a script will do it later. 

I'm not sure it's worth having a different type. Can just treat it as a hashmap?

reflectapi/src/traits.rs
```diff
@@ -728,4 +730,53 @@ mod json {
...
+            )
+        }
+    }
+
+    fn reflectapi_type_json_map(schema: &mut crate::Typespace) -> String {
+        let type_name = "serde_json::Map";
+        if schema.reserve_type(type_name) {
+            let type_def = crate::Primitive::new(
+                type_name.into(),
+                "JSON object type".into(),
```
avkonst, line 771: shall we be more specific here in the documentation.?

andyyu2004 (approved):

PR #67, opened 2024-11-05
Minimal tracing support

andyyu2004:
Add client spans and support `request-middleware::ClientWithMiddleware`.

### DIFF
--- a/reflectapi-cli/src/main.rs
+++ b/reflectapi-cli/src/main.rs
@@ -45,11 +45,17 @@ enum Commands {
         #[arg(short, long, value_delimiter = ',')]
         exclude_tags: Vec<String>,
 
+        /// Typecheck the generated code
         #[arg(short, long, default_value = "false")]
         typecheck: bool,
 
+        /// Format the generated code
         #[arg(short, long, default_value = "true")]
         format: bool,
+
+        /// Instrument the generated code with tracing
+        #[arg(short, long, default_value = "false")]
+        instrument: bool,
     },
     /// Documentation subcommands
     #[command(subcommand)]
@@ -114,6 +120,7 @@ fn main() -> anyhow::Result<()> {
             exclude_tags,
             typecheck,
             format,
+            instrument,
         } => {
             let include_tags = BTreeSet::from_iter(include_tags);
             let exclude_tags = BTreeSet::from_iter(exclude_tags);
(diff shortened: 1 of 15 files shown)


reflectapi/Cargo.toml
```diff
@@ -53,6 +53,7 @@ check_keyword = { version = "0.2.0", optional = true }
 
 # optional 3rd party dependencies for client runtime
 reqwest = { version = "0.12", optional = true }
+reqwest-middleware = { version = "0.3", optional = true }
```
hardbyte, line 56: Don't you have to add this as a feature as well?

I assume the clients will have to depend on `reqwest-tracing`, and create their custom reqwest client - might be worth updating the demo to show how to use it.

reflectapi/src/rt.rs
```diff
@@ -188,23 +191,28 @@ impl Client for reqwest::Client {
...
+#[cfg(feature = "reqwest-middleware")]
+impl Client for reqwest_middleware::ClientWithMiddleware {
+    type Error = reqwest_middleware::Error;
+
+    async fn request(
+        &self,
+        path: Url,
+        body: bytes::Bytes,
+        headers: http::HeaderMap,
+    ) -> Result<(http::StatusCode, bytes::Bytes), Self::Error> {
```
hardbyte, line 212: Nice

reflectapi-cli/src/main.rs
```diff
@@ -86,7 +86,9 @@ fn generate(
 
     let config = reflectapi::codegen::Config {
         format: true,
-        typecheck: true,
+        // Rust typechecking is a bit limited.
+        typecheck: matches!(language, Language::Typescript),
+        tracing: true,
```
hardbyte, line 91: Nice touch making the `#instrument` annotations opt in.

hardbyte (approved):

reflectapi/Cargo.toml
andyyu2004: Yep pretty much. Can do. I think it implicitly creates a feature under the same name. Not sure if that method is deprecated or not.

reflectapi/src/codegen/rust-dependency-stubs/http.rs
```diff
@@ -17,3 +18,56 @@ impl StatusCode {
...
+    pub fn insert(&mut self, _name: HeaderName, _value: T) {
+        unimplemented!()
+    }
+}
+
+pub struct HeaderName(());
+
+impl HeaderName {
+    pub fn from_bytes(_bytes: &[u8]) -> Result<Self, InvalidHeaderName> {
+        unimplemented!()
```
avkonst, line 41: it is not real code? only stubs?

andyyu2004: Yeah, it's a bit of a hack to do typechecking to ensure the generated rust code is correct for tests at least. I don't mind removing it but it's pretty helpful.

PR #68, opened 2024-11-12
feat(openapi): document error types

andyyu2004:
Taking easy approach for now and using `default` case. We could be smarter and document status code by status code but would require maintaining more information in `reflectapi.json` than we currently have.


avkonst:
approved

PR #69, opened 2024-11-13
Subcommand to serve documentation for `reflectapi.json`

andyyu2004:


reflectapi-cli/src/main.rs
```diff
@@ -64,6 +74,28 @@ fn main() -> anyhow::Result<()> {
...
+
+            let schema: reflectapi::Schema = serde_json::from_reader(std::fs::File::open(&path)?)
+                .context(
+                "Failed to parse schema file as JSON into reflectapi::Schema object",
+            )?;
+
+            let addr = format!("0.0.0.0:{port}");
+            eprintln!("Serving {} on http://{addr}", path.display());
+            let openapi = reflectapi::codegen::openapi::Spec::from(&schema);
+            rouille::start_server(addr, move |request| {
```
avkonst, line 91: why do not you stick with axum server here? we already have this dependency to use..

andyyu2004: It's not a dependency of the CLI though.

andyyu2004: Trying to avoid axum and all the async baggage for something this simple.

PR #74, opened 2024-12-05
codegen(openapi): handle date/time cases

andyyu2004:


reflectapi-demo/src/lib.rs
```diff
@@ -91,6 +91,8 @@ mod model {
         #[serde(default, skip_serializing_if = "Option::is_none")]
         #[deprecated(note = "test deprecation")]
         pub age: Option<u8>,
+        #[serde(default = "super::some_datetime")]
```
avkonst, line 94: + skip_serializing_if?

andyyu2004: If what though? It will always exist on output.

andyyu2004:
Looks like there is a bug in ts codegen outputting `std.string.String`. Will investigate

PR #77, opened 2025-02-23
Updates Axum to v0.8

hardbyte:
https://tokio.rs/blog/2025-01-01-announcing-axum-0-8-0


andyyu2004:
Thanks, I can release this under a new major version (0.x+1.y). Also spotted this in their blog which looks like the most agreeable approach to an ORM in rust I've seen so far. https://github.com/tokio-rs/toasty

andyyu2004:
Will be in 0.14.0

Issue #88, opened 2025-07-24
Expose transport metadata and raw transport in clients


hardbyte:
The generated clients have very limited access to the underlying transport layer (i.e., HTTP). This restricts developers from implementing common and important client-side logic - accessing timing info, headers  (eg, X-Rate-Limit-Remaining,  ETag).

andyyu2004:
Can this not be done at the axum layer with middleware, what's an example usecase?

avkonst:
I am not convinced this is aligned with the idea of the simple client. Why
does a client need to access something what is not in the signature?

hardbyte:
> Why does a client need to access something what is not in the signature?

To make the client easier to test and troubleshoot.

> I am not convinced this is aligned with the idea of the simple client. 

I agree that the primary goal is to provide a "simple client" where the developer doesn't have to think about the underlying transport. For the 80% use case, the interaction should feel like a simple function call, not an HTTP request. My proposed design in #89 is trying to preserve this simplicity for the common path while also adding an opt-in mechanism.

My motivation for this proposal came from writing tests for a TypeScript client. I wanted to assert that a specific API call returned the correct HTTP status code. This felt like a very reasonable and basic requirement for testing an API client. This information was completely hidden by the abstraction, same with other information such as headers.

Of course we could have exposed the status code similar to Rust to solve that immediate problem. It seemed the right time to propose a standard way to access the transport-level data consistently.

avkonst:
> For the 80% use case
The goal is to make the client applicable in 100% and we achieved it so far.

> To make the client easier to test and troubleshoot.
The proposed changes make it easier to bind to internals for the business logic. If a client does so and we do a change on internals, we might break the business logic. I fundamentally disagree with exposing internals (it is similar to exposing what blob was sent to the storage driver, when a `libc` client called `fopen`). Below I outlined how to deal with your specific need in tests and the general design principles for more generic needs.

> assert that a specific API call returned the correct HTTP status code. 
In your particular case, if you did true blackbox testing, you would only care about error handling the reflectapi provides. And it does differentiate between 
* success result as per the result signature (200 always, but a user does not need to know about the exact status), 
* structured application error as per the error type (most 4XX codes except a few but a user does not need to know the exact status code), 
* other error (network error, protocol error, like bad request, or an internal error 5XX - I believe the error context includes the underlying cause in the error for very much complete troubleshooting).
[...]

PR #89, opened 2025-07-24
Feature/expose transport layer

hardbyte:
### Why This Change Was Made

As per #88,  our previous clients offered inconsistent and minimal transport-layer information. This PR introduces a system to expose rich metadata (HTTP status, headers, timing) for both success and error responses across all generated clients. This enables advanced use cases like client-side caching, rate-limiting, and performance monitoring, while also improving the debugging experience.

A key goal was to provide this new functionality to TypeScript users **without introducing breaking changes**.

#### Rust API (Breaking Change)

Successful responses are now wrapped in an `ApiResult<T>` struct, which bundles the deserialized `value` with its `metadata`.

```rust
// Before
let pet: Pet = client.get_pet(1)?.unwrap();

// After
let result: ApiResult<Pet> = client.get_pet(1)?.unwrap();
let pet: Pet = result.value;
let metadata: &TransportMetadata = &result.metadata;

// Deref allows for ergonomic field access on the wrapped value
println!("Pet name: {}", result.name); // Works due to Deref
```

Error handling is also enhanced, with all `Error` variants now containing metadata accessible via `.transport_metadata()`.

#### TypeScript API (Fully Backward-Compatible)
For Typescript we were using a custom `Result` class so I added the metadata to that so it should be fully backward-compatible. 

**Success Handling:**
`unwrap_ok()` continues to return the value directly. The new `metadata` is available as an optional property on the `Result` object itself.

```typescript
// Existing code works without any changes:
const pet = result.unwrap_ok();

// New: Code that needs metadata can access it like this:
if (result.is_ok()) {
  const metadata = result.metadata;
  console.log(`Request succeeded with status ${metadata.status}`);
}
```

**Error Handling:**
This remains 100% backward-compatible. The existing `.status()` method on the `Err` class still works, but you can now get the full metadata object via the new `.transport_metadata()` method.

```typescript
// Existing code works without any changes:
if (result.is_err()) {
  const status = result.err().status();
}

// New: Accessing the full metadata object
if (result.is_err()) {
  const metadata = result.err().transport_metadata();
  console.log(`Request failed with headers:`, metadata.headers);
}
```

### How It Was Implemented

1.  **Core Rust Runtime (`rt.rs`):** The foundational Rust structs (`TransportResponse`, `TransportMetadata`, `ApiResult`) were created to establish a canonical model. The `Client` trait was updated to return these rich types.
2.  **TypeScript Runtime (`lib.ts`):** An optional `metadata` property was added directly to the `Result` class.
3.  **Code Generation:** The Rust and TypeScript generators were updated to accommodate these new structures and produce the APIs described above.
4.  **Testing:** The Rust and TypeScript test suites have been expanded to validate metadata content (status, headers, timing) on all success and error paths, with specific tests confirming TypeScript's backward compatibility.

### Future considerations

A minor trade-off is a slight "impurity" in the type definition of the TS Result. The Result class now has an optional metadata?: TransportMetadata property. The implementation always sets this on an Ok result, but the type system can't enforce this guarantee. An alternative, more type-pure approach would be a discriminated union:

   1 type Result<T, E> =
   2   | { kind: "ok", value: T, metadata: TransportMetadata }
   3   | { kind: "err", error: E };

With this, the compiler would know that if kind is "ok", metadata is always present. The current approach was chosen to maintain the existing Result class API.


avkonst:
> This enables advanced use cases like client-side caching, rate-limiting, and performance monitoring, while also improving the debugging experience.

* client-side caching would be better explicit on top of the clean client
* rate-limiting and retry is already possible within the custom provided client implementation, if a user wants to
* performance monitoring - same
* debugging experience - it might be, but I do not think it outweights the simplicity and clarity of the clean API and the original phylosophy - everything a client (from the programmer perspective) needs to know is in the typesafe strict signature

PR #90, opened 2025-07-27
Feature/python client

hardbyte:
Introduces complete Python client generation capabilities to ReflectAPI, providing feature parity with TypeScript and Rust client generators.


reflectapi-cli/src/main.rs
```diff
@@ -56,6 +56,27 @@ enum Commands {
...
+        #[arg(long, default_value = "true")]
+        python_sync: bool,
+
+        /// Generate testing utilities for Python (default: true)
+        #[arg(long, default_value = "true")]
+        python_testing: bool,
+
+        /// Base URL for the Python client (optional)
+        #[arg(long)]
+        python_base_url: Option<String>,
```
avkonst, line 79: why is it necessary?

hardbyte: Happy to remove or make more general

reflectapi/src/codegen/python.rs
```diff
@@ -0,0 +1,4117 @@
...
+    types.insert("std::result::Result".to_string(), "T | E".to_string());
+
+    // ReflectAPI specific types
+    types.insert(
+        "reflectapi::Option".to_string(),
+        "ReflectapiOption[T]".to_string(),
+    );
+    types.insert(
+        "reflectapi::Empty".to_string(),
+        "ReflectapiEmpty".to_string(),
```
JosiahBull, line 2954: I'm surprised you needed a custom type here - wouldn't this just map to `None`

JosiahBull (approved):
LGTM

reflectapi/src/codegen/python.rs
hardbyte: I think it is semantically correct. `None` might be inappropriate because:
- Empty structs serialize to `{}` in JSON, not `null`
- HTTP APIs distinguish between "no content" and JSON responses that are "empty object", and `null`.

reflectapi/src/codegen/python.rs
```diff
@@ -0,0 +1,4117 @@
...
+fn get_type_parameters(type_name: &str) -> Vec<&'static str> {
+    match type_name {
+        "std::boxed::Box" => vec!["T"],
+        "std::sync::Arc" => vec!["T"],
+        "std::rc::Rc" => vec!["T"],
+        "std::vec::Vec" => vec!["T"],
+        "std::option::Option" => vec!["T"],
+        "reflectapi::Option" => vec!["T"],
+        "std::collections::HashMap" => vec!["K", "V"],
+        "std::collections::BTreeMap" => vec!["K", "V"],
```
avkonst, line 2403: This is an example where it would be better to make as "unknown to Python" type and use the fallback type provided by the BTreeMap reflected info. Basically what if somebody provides a type called `my_super_crate::ConcurrentLockFreeHashMap<K, V>` with fallback annotated as HashMap? Will your codegen still work?

hardbyte: - known types with generics HashMap<String, i32> → dict[str, int]
- unknown types use the schema-level fallback resolution
- otherwise fallback to Any but preserve the original types by mapping into Python `Annotated[Any, "External type: ..."]`.


There is a very good chance that custom generic types are not handled perfectly.

reflectapi/src/codegen/python.rs
```diff
@@ -0,0 +1,4117 @@
...
+
+    // Define transformation rules as patterns
+    let transformations = [
+        // Remove common Rust std prefixes to make Python names cleaner
+        ("StdOption", ""),
+        ("StdVec", ""),
+        ("StdResult", "Result"),
+        ("StdTuple", "Tuple"),
+        ("StdCollectionsHashMap", "HashMap"),
+        ("StdCollectionsBTreeMap", "BTreeMap"),
```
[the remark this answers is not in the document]
hardbyte: Not really the same thing, but I've updated the codegen to use BTreeMap's so most of the Python generated is  deterministically ordered

PR #91, opened 2025-07-30
feat: Support internally tagged enums with tuple variants

hardbyte:
Noticed the openapi generator crashing on some enum variants while working on a Python client generator.

## Problem Description

The OpenAPI generator had an `unreachable!("internally tagged const?")` panic when processing internally tagged enums that contained:

1. **Unit variants** (e.g., `Bicycle` in internally tagged enums)
2. **Tuple variants with user-defined enums** (e.g., `ColoredItem(Color)` where `Color` is an enum)

### Root Cause

The issue occurred because:

1. The `internally_tag` function expected to never encounter `Schema::Const` values
2. However, `Schema::Const` is legitimately created for:
   - Unit variants in enums (returns `Schema::Const` directly)
   - Individual enum variants when processing `OneOf` schemas (each enum variant becomes a `Schema::Const`)
3. When `internally_tag` was applied to these cases, it hit the `unreachable!` macro

### Example Failing Code

```rust
#[derive(Serialize, Deserialize)]
enum Color {
    Red,
    Green, 
    Blue,
}

#[derive(Serialize, Deserialize)]
#[serde(tag = "item_type")]
enum ShopItem {
    #[serde(rename = "colored_item")]
    ColoredItem(Color), // This would crash the OpenAPI generator
    
    #[serde(rename = "gift_card")]
    GiftCard, // This would also crash (unit variant)
}
```

### Understanding Serde's Behavior

First, I investigated what serde actually produces for these patterns:

```rust
// For ColoredItem(Color::Red):
{
  "item_type": "colored_item",
  "Red": null
}

// For ColoredItem(Color::Green):
{
  "item_type": "colored_item", 
  "Green": null
}

// For GiftCard (unit variant):
{
  "item_type": "gift_card"
}
```

## Solution

Replaced the `unreachable!` in the `internally_tag` function:

```rust
Schema::Const { value, description } => {
    // For enum variants in internally tagged tuple variants, create object with discriminator + enum field
    let mut properties = BTreeMap::new();
    let mut required = BTreeSet::new();
    
    // Add the discriminator field
    properties.insert(
        tag_name.to_owned(),
        Property {
            description: String::new(),
            deprecated: false,
            schema: tag_schema.clone(),
        },
    );
    required.insert(tag_name.to_owned());
    
    // Add the enum value as a field (with null value as per serde's behavior)
    properties.insert(
        value.clone(),
        Property {
            description: String::new(),
            deprecated: false,
            schema: Inline(Schema::Flat(FlatSchema {
                description: String::new(),
                ty: Type::Null,
            })),
        },
    );
    required.insert(value.clone());
    
    Ok(Inline(Schema::Flat(FlatSchema {
        description: description.clone(),
        ty: Type::Object {
            title: String::new(),
            required,
            properties,
        },
    })))
}
```

### Before (Crashed)
```
thread panicked at 'internal error: entered unreachable code: internally tagged const?'
```

### After (Correct)
```json
{
  "oneOf": [
    {
      "type": "object",
      "required": ["Red", "item_type"],
      "properties": {
        "Red": { "type": "null" },
        "item_type": { "const": "colored_item" }
      }
    },
    {
      "type": "object", 
      "required": ["Green", "item_type"],
      "properties": {
        "Green": { "type": "null" },
        "item_type": { "const": "colored_item" }
      }
    },
    {
      "type": "object",
      "required": ["Blue", "item_type"], 
      "properties": {
        "Blue": { "type": "null" },
        "item_type": { "const": "colored_item" }
      }
    }
  ]
}
```

This represents how serde serializes `ColoredItem(Color::Red)` as `{"item_type": "colored_item", "Red": null}`.

## Test Coverage

Added test cases to prevent regression.


hardbyte:
CI was broken too because of conflicting Rust versions. I've updated the rust-toolchain file to use the more modern syntax which the github actions read.

reflectapi-demo/src/tests/enums.rs
```diff
@@ -135,3 +135,123 @@ pub enum Nums {
...
+    #[derive(reflectapi::Input, reflectapi::Output, serde::Deserialize, serde::Serialize)]
+    #[serde(tag = "type")]
+    enum E {
+        #[serde(rename = "a")]
+        A(Data),
+
+        #[serde(rename = "b")]
+        B { c: String },
+
+        #[serde(rename = "c")]
```
andyyu2004, line 158: This test doesn't seem as minimal as it could be. The only variant we care about here is `C` right?

reflectapi/src/codegen/openapi.rs
```diff
@@ -1054,7 +1054,45 @@ impl Converter<'_> {
...
+                    Property {
+                        description: String::new(),
+                        deprecated: false,
+                        schema: Inline(Schema::Flat(FlatSchema {
+                            description: String::new(),
+                            ty: Type::Null,
+                        })),
+                    },
+                );
+                required.insert(value.clone());
```
andyyu2004, line 1085: Bit strange to mark the null field as required perhaps? Unless serde actually requires it,

hardbyte: Serde actually DOES require it!

  Evidence from testing:
  1. Serde serialization produces: {"type": "a", "X": null} - the null field is always present
  2. Serde deserialization requires it: Without the field {"type": "a"}, serde fails with "invalid value: map, expected map with a single key"

  Why this happens:
  For internally tagged enums with tuple variants containing enum constants, serde treats each enum variant as a field that must be present with a null value.
   This is how serde distinguishes between A(A::X) and A(A::Y) - they serialize as {"type": "a", "X": null} vs {"type": "a", "Y": null}.


```rust
use serde::{Deserialize, Serialize};

#[derive(Debug, Serialize, Deserialize)]
enum A {
    X,
    Y,
}

#[derive(Debug, Serialize, Deserialize)]
#[serde(tag = "type")]
enum E {
    #[serde(rename = "a")]
    A(A),
}

fn main() {
    // Test what serde actually serializes for internally tagged enum with tuple variant containing enum
    let value = E::A(A::X);
    let json = serde_json::to_string_pretty(&value).unwrap();
    println!("Serialized JSON:");
    println!("{}", json);
    
    // Test deserialization - does it require the null field?
    println!("\nTesting deserialization:");
    
    // Test 1: With explicit null field
    let json_with_null = r#"{"type": "a", "X": null}"#;
    match serde_json::from_str::<E>(json_with_null) {
        Ok(v) => println!("✅ With null field: {:?}", v),
        Err(e) => println!("❌ With null field: {}", e),
    }
    
    // Test 2: Without the null field
    let json_without_null = r#"{"type": "a"}"#;
    match serde_json::from_str::<E>(json_without_null) {
        Ok(v) => println!("✅ Without null field: {:?}", v),
        Err(e) => println!("❌ Without null field: {}", e),
    }
}
```

     Running `target/debug/test_serde`
Serialized JSON:
{
  "type": "a",
  "X": null
}

Testing deserialization:
✅ With null field: A(X)
❌ Without null field: invalid value: map, expected map with a single key

hardbyte:
I've made the tests much more minimal and unsemantic.

reflectapi-demo/src/tests/snapshots/reflectapi_demo__tests__enums__internally_tagged_enum_with_tuple_variants-2.snap
```diff
@@ -0,0 +1,56 @@
...
+   */
+  export interface Infallible {}
+}
+
+export namespace reflectapi_demo {
+  export namespace tests {
+    export namespace enums {
+      export type A = "X" | "Y";
+
+      export type E = { type: "a" } & reflectapi_demo.tests.enums.A;
```
andyyu2004, line 40: This doesn't look right, maybe uncovered a ts codegen bug too

reflectapi/src/codegen/openapi.rs
andyyu2004: yeah makes sense. This is a super strange serialization format though. I'm not actually convinced we should support it, it seems like you'd be better off picking another enum representation. @avkonst ? The other argument is we should support whatever serde supports but this seems like a very easy to misuse case.

avkonst:
> The other argument is we should support whatever serde supports but this seems like a very easy to misuse case.

The first thing we need to do is to make sure reflect-derive macro produces structured compiler error when encounters inner tagged enum variant for non object variant data (i.e. includes no data, strings, numbers, booleans, arrays).

The second thing might be worth checking what msgpack does on this front? So we are a bit more aware not and in the future

The third thing is to collect the requirements of the intended use case. Where does it come from? What does a user want to achieve by defining such an API? When we know the implications of end user needs, we might be able to make a decision to open support for it, but until then I would rather stop on the first action above. I agree with Andy the generated format is not obvious and may lead to confusion.

If we decide to support it at the reflect API level, Partly's standard should prohibit it and the validator to trap and block it automatically.

andyyu2004:
Going to merge this for now. The following case is useful, one more ts codegen bug to fix.
```rust
    #[derive(reflectapi::Input, reflectapi::Output, serde::Deserialize, serde::Serialize)]
    #[serde(tag = "type")]
    enum A {
        X,
        Y(()),
    }
```

PR #107, opened 2025-11-03
SSE based streaming

andyyu2004:


reflectapi-schema/src/lib.rs
```diff
@@ -527,6 +522,42 @@ impl Function {
     }
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "output_kind", rename_all = "snake_case")]
+pub enum OutputType {
```
avkonst, line 527: do we really need this? seems like handling partial case of what TypeReference can already do anyway. TypeReference to Stream with ItemType and ItemErrorType would have exactly this information anyway.

reflectapi/src/builder/mod.rs
```diff
@@ -140,10 +144,10 @@ where
         F: Fn(S, I, H) -> Fut + Send + Sync + Copy + 'static,
         Fut: std::future::Future<Output = R> + Send + 'static,
         R: IntoResult<O, E> + 'static,
-        I: crate::Input + serde::de::DeserializeOwned + Send + 'static,
-        H: crate::Input + serde::de::DeserializeOwned + Send + 'static,
-        O: crate::Output + serde::ser::Serialize + Send + 'static,
-        E: crate::Output + serde::ser::Serialize + crate::StatusCode + Send + 'static,
+        I: Input + DeserializeOwned + Send + 'static,
```
avkonst, line 147: I prefer full paths as it is clear what is what in the reviews

reflectapi-schema/src/lib.rs
andyyu2004: That seems like a harder and worse way of implementing this. It should be a first class concept I think rather than some magic type.

reflectapi/src/builder/mod.rs
```diff
@@ -155,6 +159,35 @@ where
         self
     }
 
+    /// Adds a stream route to the API.
+    ///
+    /// This method takes a stream handler function and a closure that configures the
+    /// route's metadata (like its name, path, and description) using a [`RouteBuilder`].
+    pub fn stream_route<F, St, R, I, O, E1, E2, H>(
```
avkonst: where are 2 errors coming from? I think we can assume that notifications item is always a success type without an error part.

reflectapi/src/builder/mod.rs
andyyu2004: There is no ambiguity here though right? Input/Output are/StatusCode our own traits, Serialize/Deserialize are effectively known across the entire ecosystem. It significantly reduces line noise to not have it qualified.

reflectapi/src/builder/mod.rs
andyyu2004: That's too strict though, all IO is fallible, how to handle this nicely without the inner error type? It's optional anyway so it's not forced similar to how existing handlers work. `()` or `Ok(())` are valid return types due to `IntoResult`

reflectapi/src/codegen/rust-dependency-stubs/futures_util.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    impl<Item> super::Stream for Unfold<Item> {
+        type Item = Item;
+    }
+
+    impl<Item> Unpin for Unfold<Item> {}
+}
+
+pub struct Map<Item> {
```
avkonst, line 40: could you please remind me why need all of this?

reflectapi/src/codegen/lib.ts
```diff
@@ -215,6 +216,139 @@ export function __request<I, H, O, E>(
...
+export async function __stream_request<I, H, O, E>(
+  client: Client,
+  path: string,
+  input: I | undefined,
+  headers: H | undefined,
+  options?: RequestOptions,
+): Promise<Result<AsyncIterable<O>, Err<E>>> {
+  let hdrs: Record<string, string> = {
+    "content-type": "application/json",
+    "accept": "text/event-stream",
```
avkonst, line 228: that means we also serialize notifications as json? I thought reflectapi is agnostic to encoding between msgpack vs json. Previously the rule was: the server responds to the same encoding / packing as the request was. I suggest we keep following this rule.

reflectapi/src/codegen/rust-dependency-stubs/futures_util.rs
andyyu2004: can test the generated rust code typechecks. Need type definitions of dependencies to typecheck against.

reflectapi/src/codegen/lib.ts
```diff
@@ -215,6 +216,139 @@ export function __request<I, H, O, E>(
...
+      }
+    }
+  } catch (e) {
+    return new Result<AsyncIterable<Result<O, IE>>, Err<E>>({
+      err: new Err({ other_err: e }),
+    });
+  }
+}
+
+async function* __sse_to_async_iterable<O>(
```
avkonst, line 264: Interesting limitation we need to be aware of. Is our API gateway https 2 capable? 

Image

reflectapi/src/codegen/lib.ts
andyyu2004: I don't think any of the clients actually support messagepack right now. That can be future work I think

reflectapi/src/codegen/lib.ts
```diff
@@ -215,6 +216,139 @@ export function __request<I, H, O, E>(
...
+  response: Response,
+  options?: RequestOptions,
+): AsyncIterable<O> {
+  const body = response.body;
+  if (!body) return;
+  const reader = body.pipeThrough(new TextDecoderStream()).pipeThrough(new __EventSourceParserStream()).getReader();
+  try {
+    while (true) {
+      if (options?.signal?.aborted) break;
+      const { done, value } = await reader.read();
```
avkonst, line 274: does this give complete message separated by 2 new lines? or it can cut it in the middle as pure http/tcp stream?

reflectapi/src/codegen/lib.ts
```diff
@@ -215,6 +216,139 @@ export function __request<I, H, O, E>(
...
+      if (done) break;
+      yield JSON.parse(value.data) as O;
```
avkonst, line 276: as far as I can see the protocol allows event type prefix (event: <string>). I suggest we ignore it, and hardcode on the server side to just `notification`

reflectapi/src/codegen/lib.ts
```diff
@@ -215,6 +216,139 @@ export function __request<I, H, O, E>(
...
+  const body = response.body;
+  if (!body) return;
+  const reader = body.pipeThrough(new TextDecoderStream()).pipeThrough(new __EventSourceParserStream()).getReader();
+  try {
+    while (true) {
+      if (options?.signal?.aborted) break;
+      const { done, value } = await reader.read();
+      if (done) break;
+      const parsed = JSON.parse(value.data);
+      if ("__reflectapi_reserved_stream_error" in parsed) {
```
avkonst: how can that happen? I think only network error types are possible here, and we assume the server does not send errors in the event stream. Only the original response error needs to be handled.

reflectapi/src/codegen/python.rs
```diff
@@ -2974,10 +2977,12 @@ fn render_function(
...
-    let output_type = if let Some(output_type) = function.output_type.as_ref() {
-        type_ref_to_python_type_simple(output_type, schema, implemented_types, &[])?
-    } else {
-        "Any".to_string()
+    let output_type = match &function.output_type {
+        OutputType::Complete {
+            output_type: Some(output_type),
+        } => type_ref_to_python_type_simple(output_type, schema, implemented_types, &[])?,
+        OutputType::Complete { output_type: None } => "Any".to_string(),
+        OutputType::Stream { .. } => unreachable!("stream endpoints should be filtered out"),
```
avkonst, line 2985: do you mean python client will not support these?

reflectapi/src/codegen/lib.ts
andyyu2004: Also sse requires utf8 so we'd have to base64 or something ontop of messagepack

reflectapi/src/codegen/lib.ts
andyyu2004: Well I implemented it such that it supports per event errors, and I still think that's a useful thing to support.

reflectapi/src/codegen/python.rs
andyyu2004: Will leave to @hardbyte if it's necessary

reflectapi/src/codegen/rust.rs
```diff
@@ -834,10 +845,89 @@ mod templates {
...
+                    out.push_str("        #[deprecated]");
+                } else {
+                    write!(out, "        #[deprecated(note = \"{deprecation_note}\")]").unwrap();
+                }
+            }
+            if let Some(item_error_type) = &self.item_error_type {
+                write!(
+                    out,
+                    "        {}{}pub async fn {}(&self, input: {}, headers: {})\n\
+                         -> reflectapi::rt::FallibleStreamResponse<{}, {}, {}, C::Error>\n\
```
avkonst: why is it not the same Response type like in Complete case, but with generic parameter set to a Stream<Item>? Can we go without extra customer response types here?

reflectapi/src/codegen/rust.rs
```diff
@@ -834,10 +845,89 @@ mod templates {
...
+                    item_error_type,
+                    self.error_type,
+                    self.path,
+                )
+                .unwrap();
+            } else {
+                write!(
+                    out,
+                    "        {}{}pub async fn {}(&self, input: {}, headers: {})\n\
+                         -> reflectapi::rt::StreamResponse<{}, {}, C::Error>\n\
```
avkonst: Same question here

reflectapi/src/codegen/typescript.rs
```diff
@@ -622,6 +622,51 @@ mod templates {
...
+        pub item_error_type: Option<String>,
+        pub error_type: String,
+    }
+
+    impl StreamFunctionImplementationTemplate {
+        pub fn render(&self) -> String {
+            if let Some(item_error_type) = &self.item_error_type {
+                format!(
+                    "function {name}(client: Client) {{\n\
+                         return (input: {input_type}, headers: {input_headers}, options?: RequestOptions) => __stream_request_fallible<\n\
```
avkonst: same question here, if we could reuse the original normal Response type instead of introducing new one

reflectapi/src/codegen/rust.rs
andyyu2004: As in just inline the type alias? It's just huge and long and noisy

reflectapi/src/codegen/rust.rs
andyyu2004: same answer

reflectapi/src/rt.rs
```diff
@@ -68,11 +87,19 @@ impl<AE: std::error::Error + 'static, NE: std::error::Error + 'static> std::erro
     }
 }
 
+pub type BoxStream<T> = Pin<Box<dyn Stream<Item = T> + Send + 'static>>;
+
+pub type StreamResponse<T, AE, NE> = Result<BoxStream<Result<T, Error<AE, NE>>>, Error<AE, NE>>;
+
+pub type FallibleStreamResponse<T, IE, AE, NE> =
```
avkonst: All these IE, AE, NE are hard to read and understand what are these for. Either need comments or proper longer names for generic params. Also consider if we can cut back complexity by assuming errors are not coming from the server, only network disconnect to be considered.

reflectapi/src/codegen/typescript.rs
andyyu2004: Isn't this already what you want? 

```
    get_first: (
      input: {},
      headers: myapi.proto.Headers,
      options?: RequestOptions,
    ) => AsyncResult<
      myapi.model.output.Pet | null,
      myapi.proto.UnauthorizedError
    >;
    /**
     * Stream of change data capture events for pets
     */
    cdc_events: (
      input: {},
      headers: myapi.proto.Headers,
      options?: RequestOptions,
    ) => AsyncResult<
      AsyncIterable<myapi.model.output.Pet>,
      myapi.proto.UnauthorizedError
    >;
```

reflectapi/src/rt.rs
andyyu2004: Please respond to one of the questions about how the server should handle errors while processing the stream.

reflectapi/src/rt.rs
```diff
@@ -193,6 +226,254 @@ where
...
+fn __serialize_headers_for_stream<H: serde::Serialize>(
+    headers: H,
+) -> Result<http::HeaderMap, (String, ProtocolErrorStage)> {
+    let headers = serde_json::to_value(&headers)
+        .map_err(|e| (e.to_string(), ProtocolErrorStage::SerializeRequestHeaders))?;
+
+    let mut header_map = http::HeaderMap::new();
+    header_map.insert(
+        http::header::ACCEPT,
+        http::HeaderValue::from_static("text/event-stream"),
```
avkonst, line 228: are we definitely bound to require this accept header? might be easier to assume that header by default? also we can assume the server only outputs new line (or zero byte in case of msgpack) separated json lines?

reflectapi/Cargo.toml
```diff
@@ -25,6 +25,7 @@ reflectapi-schema = { path = "../reflectapi-schema", version = "0.16.10" }
 
 # mandatory 3rd party dependencies
 serde = { version = "1.0.197", features = ["derive"] }
+futures-util = "0.3"
```
avkonst, line 28: that is optional, is not it? not always to be required

andyyu2004: We can feature flag it behind sse if you really want, but anyone who uses async will have this in tree anyway

reflectapi-schema/src/lib.rs
avkonst: That would not be a magic type right? Stream type is known in Rust? In this case you two ways of representing the same thing, which is not ideal. But ok, I can leave with explicit in enum

reflectapi/src/builder/mod.rs
avkonst: well, ok. but harder for review. I review a lot and always feel easier to review when full paths. You can leave it then

reflectapi/src/builder/mod.rs
avkonst: What would a client do with a notification, which says an error? In any case it would be mapped to internal error, it would require to resend a request.. Simpler flow (despite being stricter) is much easier to implement and handle for the client:
* send a request
* handle success subscription or error (eg. not authorized)
* if success receive notifications (of a single type)
* then either a client stops receiving (drops a receiver) or a channel is ended (server aborts or network aborts)
* a client reconnects repeating from the above step
Where is the above not enough? Do you have a usecase where a client needs to receive errors in notifications? What would these notifications mean?

reflectapi/src/codegen/lib.ts
avkonst: > I don't think any of the clients actually support messagepack right now. 
I think Rust did.

avkonst: > Also sse requires utf8 so we'd have to base64 or something ontop of messagepack

Yeah, do we have to conform here to text? We do not have too. We can just send line (or zero byte) separated messages.

reflectapi/src/codegen/lib.ts
avkonst: that is important question

reflectapi/src/rt.rs
```diff
@@ -193,6 +226,254 @@ where
...
+                    }
+                    continue;
+                }
+
+                match body.next().await {
+                    Some(Ok(chunk)) => {
+                        buffer.push_str(&String::from_utf8_lossy(&chunk));
+                    }
+                    Some(Err(_)) => {
+                        return Some((Err("stream error".to_string()), (body, buffer, data)));
```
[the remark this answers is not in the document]
avkonst: yeah, same question I asked about incomplete messaging recieved

reflectapi/src/codegen/typescript.rs
avkonst: yes, it is. Looks good.
All we want

reflectapi/src/rt.rs
avkonst: what errors are there coming to the server on the server side?
all I can think of would map to an internal error, there is no way for a user originated error to come here. No way a client receiving notification can do anything about it. Best a server can do is abort the connection due to an internal error. If you want it a bit more explicit, maybe send last frame as a special marker of such an internal error event before aborting the connection, and a client would report an internal error and reconnect.

reflectapi/Cargo.toml
avkonst: I think following the feature flag at cargo level is the best option and aligns with everything else we have

reflectapi/src/codegen/lib.ts
andyyu2004: That's no longer SSE, but is an option I suppose. jsonl/ndjson streaming. 0 byte doesn't work with message pack right, msgpack can contain 0 bytes anywhere within.

reflectapi-schema/src/lib.rs
andyyu2004: Stream is not a type though, it's a trait. It may be doable somehow, but unclear. I think this is a better approach.

reflectapi/src/codegen/lib.ts
andyyu2004: Which bit of code are you referring to in particular?

reflectapi/src/builder/mod.rs
andyyu2004: What about on the server side? How would you handle failure there (probably some form of internal error).

reflectapi/src/codegen/lib.ts
andyyu2004: We're not using the EventSource API since it doesn't handle auth so don't think we run into this?

reflectapi/src/codegen/typescript.rs
andyyu2004: Ok, was fine already then. Resolving

reflectapi/src/rt.rs
andyyu2004: I'll just remove the failure case then, can be added on later if required

reflectapi/src/rt.rs
andyyu2004: SSE is a specific format, we can support different formats in the future based on the header if we really want to.

reflectapi/src/rt.rs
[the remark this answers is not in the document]
andyyu2004: yeah I think the rust one is broken, the typescript implementation should be fine. I should just find some good sse client library.

reflectapi/src/codegen/lib.ts
andyyu2004: not sure what you're referring to here, I think we only send data: .  The ts sse client is vendored, easiest to not touch it.

reflectapi/src/rt.rs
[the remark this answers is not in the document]
andyyu2004: Replaced with sseer. Not any very good implementation of it unfortunately, this one seems good enough

reflectapi/Cargo.toml
andyyu2004: Bit hard to get rid of this one entirely, also zero value. Not worth trying in my opinion

andyyu2004: The sse support crate is feature flagged.

reflectapi/src/codegen/lib.ts
avkonst: I am asking in general. Once await for reading is complete will it have full message from the server or can get cut half as partially received tcp bytes?

reflectapi/src/rt.rs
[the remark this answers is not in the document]
avkonst: this is a bit strange there is need in the library. It is pretty much dealing with bytestream of response body. Need to poll for arriving chunks of bytes until a complete message can be reconstructed, then notify and poll for the next one. Maybe not even a library required here.

reflectapi/src/rt.rs
avkonst: uhh... ok

reflectapi/src/rt.rs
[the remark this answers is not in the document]
andyyu2004: As in you want to implement sse parsing ourselves?

reflectapi/src/codegen/lib.ts
andyyu2004: This implementation will work correctly.

reflectapi/src/rt.rs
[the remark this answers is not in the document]
avkonst: I thought that would not be difficult, provided there are messages coming in split by well known separator. 
Very typical task for data receivers over a stream

avkonst:
@andyyu2004 checking how far is it from being possible to merge?

andyyu2004:
I think it's basically mergeable as MVP and for use in experimental endpoints.

andyyu2004:
In particular, I wanted feedback from Eli on the js api

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
+
+    if (streamResult.is_err()) {
+        console.log('stream error:', streamResult.unwrap_err().toString());
+        return;
+    }
+
+    const received: string[] = [];
+    const streamDone = (async () => {
+        try {
+            for await (const pet of streamResult.unwrap_ok()) {
```
avkonst: ok, looks beautiful. is it a generator? what does unwrap_ok() do here?

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
-    console.log('removing pet');
-    (
-        await c.pets.remove({
-            name: 'Bobby',
-        }, {
-            authorization: 'password'
-        })
-    ).unwrap_ok();
+    console.log('streaming cdc events while mutating pets');
+    const controller = new AbortController();
```
avkonst, line 49: where does the AbortController come from? how is it linked with signal below? any description how it works? not very obvious from the code

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
+    }
+
+    const received: string[] = [];
+    const streamDone = (async () => {
+        try {
+            for await (const pet of streamResult.unwrap_ok()) {
+                console.log('received event:', pet.name, JSON.stringify(pet.kind));
+                received.push(pet.name);
+            }
+        } catch (e: any) {
```
avkonst: how to handle without catch? is there non throwing alternative to unwrap_ok()?

avkonst: is it possible to handle by a stricter type, instead of doing text match on error name below?

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
+            if (e.name !== 'AbortError') throw e;
+        }
+    })();
+
+    await new Promise((r) => setTimeout(r, 100));
+
+    console.log('creating Whiskers');
+    (await c.pets.create({
+        name: 'Whiskers',
+        kind: { type: 'cat', lives: 9 },
```
avkonst, line 72: 9 lives :)

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
-            authorization: 'password'
-        })
-    ).unwrap_ok();
+    console.log('streaming cdc events while mutating pets');
+    const controller = new AbortController();
+    const streamResult = await c.pets.cdc_events({}, {
+        authorization: 'password'
+    }, { signal: controller.signal });
+
+    if (streamResult.is_err()) {
```
avkonst, line 54: and it differentiates between app error and other error right? like normal API? and it is possible to get to the strict type definition of the error type here?

reflectapi-demo/clients/typescript/index.ts
```diff
@@ -49,14 +45,124 @@ async function main() {
...
+            for await (const result of fallibleStreamResult.unwrap_ok()) {
+                if (result.is_ok()) {
+                    const pet = result.unwrap_ok();
+                    console.log('fallible received ok:', pet.name, JSON.stringify(pet.kind));
+                    receivedFallible.push(`ok:${pet.name}`);
+                } else {
+                    const err = result.unwrap_err();
+                    console.log('fallible received err:', err.message);
+                    receivedFallible.push(`err:${err.message}`);
+                }
```
avkonst: I think we discussed to remove fallible stream alternative as it can not fail rather than with internal server error, which we discussed would be simpler disconnect from the server side

avkonst:
@andyyu2004 I am happy with the generated TS api. I understand it is generator, so looks very clean. I left few questions on abort controller, it is not obvious what happens there, how and why.
If you answer my questions, I suggest we merge. Happy with it. Very clean API.

avkonst:
@andyyu2004 may we have a quick call in regards to this? I would like to merge it but would like to have questions answered.

PR #116, opened 2025-11-21
method to support default for converting into std option

binkynz:


andyyu2004:
I don't think the semantics of this is obvious enough or has enough precedent. Can you inline the explicit match when this logic is required?

PR #121, opened 2026-03-23
feat: Add support for `reflectapi(required)` which will force an input  to be required in the generated clients, even if it is optional/has a default. This allows for introduction of new required properties without breaking existing deployed clients.

JosiahBull:


avkonst:
@JosiahBull have you achieved the desired outcome with the transformer attribute?

JosiahBull:
It was quicker to just go with `Option<T>` instead, closing this MR now.

avkonst:
sorry, we do not want to discourage you, your usecase was legitimate and we offered you to use existing feature which allows you to achieve exactly what you wanted. That approach helped us to avoid adding a feature, if it can be achieve by the existing means. Would you still like to try the proposed approach?

PR #122, opened 2026-03-26
Andrey/refactoring cleanup

avkonst:
Remove aksama dependency. The snapshot difference is due to formatting


hardbyte:
Included in #124

PR #124, opened 2026-03-27
feat(schema): add semantic IR and symbol ID infrastructure

hardbyte:
## Summary

Adds semantic IR infrastructure (#96), fixes Python codegen for flattened tagged-union fields (#123), makes Python a first-class codegen backend with namespace classes, typed errors, field descriptions, and SemanticSchema-driven code generation.

## Semantic IR Infrastructure (`reflectapi-schema`)

| Module | Purpose |
|--------|---------|
| `symbol.rs` | `SymbolId` / `SymbolKind` — stable unique identifiers for all schema symbols |
| `ids.rs` | `ensure_symbol_ids()` — canonical ID assignment with cross-typespace disambiguation |
| `semantic.rs` | Immutable `SemanticSchema`, `SemanticType`, `SymbolTable`, `ResolvedTypeReference` |
| `normalize.rs` | `NormalizationPipeline` + `Normalizer` (`&Schema` → `SemanticSchema`) |

Schema type changes: `id: SymbolId` on all types, `Normalizer::normalize(&Schema)`, `original_name` on `SemanticType` preserving pre-normalization qualified names.

## Python Codegen — Now Driven by SemanticSchema

The Python codegen uses `SemanticSchema` as the primary driver:
- **Type iteration**: `semantic.types()` (deterministic BTreeMap order, replaces manual topological sort)
- **Import detection**: `SemanticType` pattern matching
- **Function ordering**: `semantic.functions()`
- Raw `Schema` kept for rendering (concrete field/variant data)

## Python Codegen — Wire-Compatible Flatten (#123)

For `#[serde(flatten)]` on internally-tagged enums, generates per-variant models merging parent fields + tag + variant fields into a discriminated union `RootModel`.

## Python Codegen — First-Class DX

- **Namespace classes** mirroring Rust module structure (`auth.UsersSignInRequest`)
- **Field descriptions** via `Field(description="...")`
- **Typed error returns** — `ApiResponse[OutputType, ErrorType]` in method signatures
- **Typed error deserialization** — `ApplicationError.typed_error` as Pydantic model
- **Typed list responses** — `list[Model]` via `TypeAdapter` (was returning raw dicts)
- **Fast JSON parsing** — Pydantic's Rust-based `validate_json(bytes)`
- **Factory classes removed** — direct construction via namespace types (-13% file size)
- **Docstring escaping** for backslashes, triple-quotes, Python keywords in method/parameter names

## Python Runtime Fixes

[...]

Issue #96, opened 2025-08-21
Proposal: Introduce stable Symbol IDs


hardbyte:
To evolve our code generation capabilities beyond simple templating, we can aim to build a more robust, multi-stage compiler style pipeline.
The current system relies heavily on string-based name lookups . This makes it rather fragile at circular reference detection and  dependency analysis.

This proposal introduces a foundational change to `reflectapi-schema`: the integration of a stable and unique `SymbolId` for every type, field, variant, and function in the schema. This would be the first step towards enabling a reliable pipeline architecture.

## Why

**Compiler Architecture**: A modern compiler pipeline requires a reliable way to track symbols (types, functions, etc.) across different stages of transformation (e.g., parsing -> normalization -> semantic analysis -> code generation). String names are insufficient because they can be ambiguous or change during normalization. SymbolId provides a canonical reference that remains stable throughout the entire process.

**Enabling Determinism and Stability**: By using identifiers that have a stable Ord implementation (based on kind and path), we can process schema components in a deterministic order. This guarantees that code generation is perfectly reproducible, eliminating random diffs in generated files and making version control history clean and predictable.

The true point of this change is the features it makes possible:

- Accurate Dependency Tracking: We can build a dependency graph between types, which is essential for detecting and resolving circular references.
-  Transformations: Features like generic monomorphization (generating concrete types like `List<User>` from a generic `List<T>`) become feasible because we can reliably track and substitute type symbols.
-  Improved Error Reporting: When a type reference is invalid, the normalization pipeline can report exactly which symbol failed to resolve, providing much clearer and more actionable error messages to the user.

Better and more consistent support across languages: A stable IR with unique symbol identifiers is a prerequisite for a scalable multi-language code generator. The front-end of the pipeline (parsing and normalization) can be shared across all target languages, while the back-ends (e.g., TypeScript, Rust, Python generators) can all rely on the same stable symbol information.

## How

`reflectapi-schema/src/symbol.rs`, will define the identifier:

```rust
pub struct SymbolId {
    pub kind: SymbolKind, // e.g., Struct, Enum, Field
    pub path: Vec<String>, // e.g., ["api", "User", "id"]
    pub disambiguator: u32,
}
```

The `id: SymbolId` field will be added to all core schema structs (Schema, Struct, Enum, Field, Function, etc.). The derive macros (`#[derive(ReflectAPI)]`) will have to be updated to automatically generate a preliminary, simple ID at compile time.

[...]

PR #128, opened 2026-03-29
fix(python): DX improvements — compact enums, name truncation, bug fixes

hardbyte:
## Summary

Follow-up to #124 and the review on that merge, plus the architectural follow-up from #96.

This branch keeps the Python DX fixes, then takes the larger separation-of-concerns step from the latest review:

- create a new `reflectapi-schema-codegen` crate for compiler-owned schema processing
- move `SymbolId` / `SymbolKind`, schema ID assignment, normalization, and `SemanticSchema` out of `reflectapi-schema`
- keep `reflectapi-schema` focused on the raw interchange schema and raw-schema transforms
- update Python codegen to consume `reflectapi-schema-codegen` for semantic ordering and normalization
- replace the public stdlib ID table with compiler-private external-type inference instead of maintaining a growing shared constant
- fix Python de-stuttering collisions by falling back to the original name when shortening would collide
- fix generated Python package exports so `__init__.py` re-exports `Client`
- remove placeholder enum docstrings like `Generated enum.`
- move and tighten the architecture doc in the mdBook so it reflects the current pipeline and crate split

## Why

The review feedback was directionally right: raw schema concerns and compiler/codegen concerns had become mixed together.

The important distinction is:

- `reflectapi-schema` should stay the wire/interchange layer produced by reflection and serialized to JSON
- compiler-owned identity, normalization, dependency analysis, and semantic IR should live in a separate layer
- backend-local mappings such as Python type rendering stay in the backend, not in the shared schema crates

This change moves the branch closer to that design instead of only making local Python fixes.

## Current backend shape

- Python now consumes the compiler crate plus the raw schema
- Rust still renders mostly from raw schema after backend-local preprocessing
- TypeScript still renders mostly from raw schema after backend-local preprocessing
- OpenAPI still walks raw schema directly

That leaves #129 as the follow-up for moving more backends toward a common backend contract derived from `SemanticSchema`.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check -p reflectapi-schema-codegen -p reflectapi --features codegen`
- [x] `cargo test -p reflectapi-schema-codegen -p reflectapi --features codegen --lib`
- [x] `cargo test --workspace`
- [x] `/home/brian/.cargo/bin/mdbook build` (in `docs/`)

## Notes

- The compiler crate still uses minimal private inference for well-known external types so normalization can assign stable placeholder identities without pushing backend mappings into the raw schema contract.
- The raw schema still carries `LanguageSpecificTypeCodegenConfig`, but today that is only the serialized Rust metadata (`additional_derives`). Python mappings are backend-local again.


reflectapi/src/traits.rs
```diff
@@ -63,11 +63,180 @@ pub(crate) fn reflectapi_type_empty(
...
+        "std::net::Ipv6Addr" => Some(python_type_codegen_config_with(
+            "IPv6Address",
+            &["from ipaddress import IPv6Address"],
+            &[],
+            false,
+            false,
+        )),
+        _ => None,
+    }
+}
```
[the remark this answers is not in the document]
avkonst: This is especially unclear why we inject it into the schema, when it is really codegen time problem not the problem of reflection or extra annotation.
Rust language specific option (additional derives to add to the types) can be parsed additionally from code - this is what would make sense to include into the schema because there is no other source for it and it is extracted from source code. Derive macro does not have any extra specific attributres for Python, like for Rust, so it is not part of the schema.  These code lines of mapping are language specific, and static for codegen time.

avkonst (changes requested):

avkonst:
I reviewed the latest diff. Seems like fixed along the lines we discussed. 
Only unclear is how the symbol id is assigned from Default::default()

hardbyte:
> I reviewed the latest diff. Seems like fixed along the lines we discussed. Only unclear is how the symbol id is assigned from Default::default()

The `Default::default()` pattern is a bit of shortcut/pragmatic workaround. IDs are assigned lazily by `ensure_symbol_ids()` during normalization, but the id field lives on the raw schema structs where it's not really needed. As #129 notes, this is a case of compiler concerns leaking into the interchange layer. The cleaner long-term design would be to move symbol identity out of raw Schema and into the compiler-owned layer - probably on `SemanticSchema` types (which already exist). That would eliminate the `Default::default()` pattern entirely. I'll see if I can add that in this PR too...

hardbyte:
@avkonst `SymbolId` is now removed from all raw schema types (Schema, Function, Struct, Enum, Field, Variant, Primitive). IDs are now assigned later and used by the normalizer when building `SemanticSchema`.

reflectapi-schema-codegen/src/ids.rs
```diff
@@ -1,237 +1,277 @@
-/// ID assignment utilities for ensuring unique, stable SymbolIds before normalization
+/// Compiler-owned ID assignment for schema symbols.
 ///
-/// This module fixes the core issue where JSON-deserialized schemas have default
-/// SymbolIds that cause conflicts during normalization discovery.
+/// SymbolIds are a compiler concept used during normalization and codegen.
+/// They are NOT stored on raw schema types — raw types are the interchange
+/// format (JSON-serializable, derive-macro-produced). Instead, IDs are
+/// assigned here and passed to the normalizer as a side table.
```
avkonst, line 6: Ideally we have this business taken care of in a separate crate something like reflectapi-schema-codegen. There are quite a few existing tools (supportive functions for various transformations) from the reflectapi-schema could move in there.

reflectapi-schema/src/ids.rs
```diff
@@ -1,237 +1,277 @@
-/// ID assignment utilities for ensuring unique, stable SymbolIds before normalization
+/// Compiler-owned ID assignment for schema symbols.
 ///
-/// This module fixes the core issue where JSON-deserialized schemas have default
-/// SymbolIds that cause conflicts during normalization discovery.
+/// SymbolIds are a compiler concept used during normalization and codegen.
+/// They are NOT stored on raw schema types — raw types are the interchange
+/// format (JSON-serializable, derive-macro-produced). Instead, IDs are
+/// assigned here and passed to the normalizer as a side table.
 use crate::symbol::STDLIB_TYPES;
```
avkonst: that is not needed right? does not make sense to have something special for stdlib types

reflectapi-schema-codegen/src/symbol.rs
```diff
@@ -12,6 +12,7 @@ pub struct SymbolId {
     Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, serde::Serialize, serde::Deserialize,
 )]
 pub enum SymbolKind {
```
avkonst, line 14: if I understand it correctly, that would be also a candidate for reflectapi-schema-codegen

Issue #136, opened 2026-05-07
0.17.0 version breaking Client injector interface


avkonst:
The SSE work introduced the breaking API change which I did not notice:
from:
```
export interface Client {
  request(
    path: string,
    body: string,
    headers: Record<string, string>,
    options?: RequestOptions,
  ): Promise<[number, string]>;
}
```
to:
```
export interface Client {
  request(
    path: string,
    body: string,
    headers: Record<string, string>,
    options?: RequestOptions,
  ): Promise<Response>;
}
```

Our options to converge are:
1. Request => Response. Most flexible but too generic for implementors to implement, and leaky abstraction, eg. lets to hijack access to response headers, which we agreed to do type safe way via codegen. Also, complicates introspection of Requests and Responses very difficult.
```
export interface Client {
  request(
    req: Request,
  ): Promise<Response>;
}
```

2. as before but with possible `ReadableStream<string>` on return type. Restricted explicitly set expectations for implementors AND non breaking! Cons: an implementor should explicitly turn to returning the stream for enabling support for notifications, i.e. would not be trapped at compile time. Problem there is no way to know when stream is expected, when complete is expected if a client implementor needs to introspect the response data.
```
export interface Client {
  request(
    path: string,
    body: string,
    headers: Record<string, string>,
    options?: RequestOptions,
  ): Promise<[number, string | ReadableStream<string>]>;
}
```

3. as before but return stream of text, where a simple request (non streaming) would return a stream of one message and then closed stream). Breaking change, but not too much, easier to implement. Easier to hook introspection for the data. Stronger contract, not leaky.
```
export interface Client {
  request(
    path: string,
    body: string,
    headers: Record<string, string>,
    options?: RequestOptions,
  ): Promise<[number, ReadableStream<string>]>;
}
```

I think I prefer option 3.

avkonst:
@andyyu2004 FYI. If you are interested in to participate

avkonst:
cc @hardbyte

avkonst:
other thing to consider here: if we decide to turn on msgpack encoding in the future, we would prefer to deal with bytes, rather than text.

avkonst:
option 1 is also viable if no content introspection is needed by the implementor. Currently it is not needed anywhere in Partly code base. We only inject transport implementor, like playwright, for recording traces.

avkonst:
option 4 from @hardbyte , which explicitly declares what we care about.
```
export interface RequestOptions {
[...]

PR #137, opened 2026-05-07
Fix generated client transport shape

hardbyte:
## Summary
- Implements one possible solution for #136, effectively an option 4 shape using explicit generated-client transport DTOs instead of exposing generic Fetch `Request` / `Response` types.
- Updates TypeScript, Rust, and Python client runtimes to use a consistent request/response abstraction while keeping each language idiomatic.
- Replaces generated TypeScript SSE `TextDecoderStream` usage with a typed `TransformStream<Uint8Array, string>` helper so newer DOM typings validate.

## Validation
- `cargo run --bin reflectapi -- codegen --language typescript --schema reflectapi-demo/reflectapi.json --output /tmp/opencode/reflectapi-ts-typecheck --typecheck`
- `cargo test --package reflectapi-demo`
- `cargo test --package reflectapi --all-features`
- `pnpm run build` in `reflectapi-demo/clients/typescript`
- `SQLX_OFFLINE=true cargo clippy --all-targets --all-features -- -D warnings`
- `SQLX_OFFLINE=true cargo build --workspace`
- `uv run pytest tests/test_client.py` in `reflectapi-python-runtime`


reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -12,16 +12,31 @@ export interface RequestOptions {
   signal?: AbortSignal;
 }
 
+export interface ClientRequest {
```
avkonst: I would rename to just Request... because Client prefix does not make sense here.

avkonst: same for headers and response types

hardbyte: fair point - will do

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -12,16 +12,31 @@ export interface RequestOptions {
...
+  path: string;
+  method: "POST";
```
avkonst: seems like method field is not needed as everything goes via post anyway

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -279,21 +316,27 @@ export async function __stream_request<I, H, O, E>(
...
 async function* __sse_to_async_iterable<O>(
-  response: Response,
+  response: ClientResponse,
   options?: RequestOptions,
 ): AsyncIterable<O> {
   const body = response.body;
   if (!body) return;
   const reader = body
-    .pipeThrough(new TextDecoderStream())
+    .pipeThrough(__text_decoder_stream())
```
avkonst: why TextDecoderStream was not enough?

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -302,20 +345,47 @@ async function* __sse_to_async_iterable<O>(
   }
 }
 
+function __text_decoder_stream(): TransformStream<Uint8Array, string> {
+  const decoder = new TextDecoder();
+  return new TransformStream<Uint8Array, string>({
+    transform(chunk, controller) {
+      const text = decoder.decode(chunk, { stream: true });
```
avkonst: will it handle the case of a char split in the middle due to chunk split at tcp/http level?

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -302,20 +345,47 @@ async function* __sse_to_async_iterable<O>(
...
+    while (true) {
+      const { done, value } = await reader.read();
+      if (done) break;
+      body += decoder.decode(value, { stream: true });
+    }
+    body += decoder.decode();
+    return body;
+  } finally {
+    reader.releaseLock();
+  }
```
avkonst: not quite right, you need to drain the input stream to byte buffer first of all, then decode.

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -302,20 +345,47 @@ async function* __sse_to_async_iterable<O>(
...
-    options?: RequestOptions,
-  ): Promise<Response> {
-    return (globalThis as any).fetch(`${this.base}${path}`, {
-      method: "POST",
-      headers: headers,
-      body: body,
-      signal: options?.signal,
+  public request(request: ClientRequest): Promise<ClientResponse> {
+    return (globalThis as any).fetch(`${this.base}${request.path}`, {
+      method: request.method,
```
avkonst: always POST by design

reflectapi-demo/clients/typescript/generated.ts
```diff
@@ -190,9 +205,15 @@ export function __request<I, H, O, E>(
     }
   }
   return client
-    .request(path, JSON.stringify(input), hdrs, options)
+    .request({
+      path,
+      method: "POST",
+      headers: hdrs,
+      body: new TextEncoder().encode(JSON.stringify(input) ?? "{}"),
```
avkonst, line 213: why do we fallback to {}? I think more correct is zero bytes, i.e. true empty bytes array. Same question for stream version

reflectapi-demo/clients/typescript/generated.ts
hardbyte: I'll rename across all three languages. Rust’s `Request` currently carries `url: Url`, while TS and Python carry path: string. Should we align Rust to `path: String` so all three transports look the same shape?

reflectapi-demo/clients/typescript/generated.ts
hardbyte: I keep forgetting this library only supports POST. Will remove

reflectapi-demo/clients/typescript/generated.ts
hardbyte: I'll try native `TextDecoderStream()` again - it was just failing typecheck but I could have had an old TS/lib.dom.

reflectapi-demo/clients/typescript/generated.ts
hardbyte: It would yes - but will go away if we can adopt to TextDecoderStream anyway which would be nicer.

reflectapi-demo/clients/typescript/generated.ts
hardbyte: I don't believe that's accurate. But for a non-streaming body we can hopefully delegate to the platform anyway with something like:

```ts
  if (!response.body) return "";
  return await new (globalThis as any).Response(response.body).text();
  ```

then it would do the decoding dance

reflectapi-demo/clients/typescript/generated.ts
hardbyte: Yeah that's not ideal. Asseme we should send  zero-byte body and skip the Content-Type: application/json header.

hardbyte: I don't want to change that behavior in this pr though so I'll just try clean this up.

avkonst: >  and skip the Content-Type: application/json header.

technically we do not need to set this header on request and I am sure we do not set it on response. It is assumed default by the server in our case, I think.

avkonst: btw, even if the content type header is set, it does not influence the body to be empty, because it the length of it is controlled by another header

reflectapi-demo/clients/typescript/generated.ts
avkonst: > I keep forgetting this library only supports POST

your type declaration explicitly says method can be only equal to "POST"

reflectapi-demo/clients/typescript/generated.ts
avkonst: yes, I agree to align

reflectapi-demo/clients/typescript/generated.ts
hardbyte: Switched back to native TextDecoderStream() with a single explicit cast (`as unknown as ReadableWritablePair<string, Uint8Array>`) at the pipeThrough site, with a comment explaining why.

reflectapi-demo/clients/typescript/generated.ts
hardbyte: ok so for endpoints with no input type, send a true zero-byte body and skip the Content-Type header entirely.
The fallback ?? "{}" then becomes dead code and goes away.

reflectapi-demo/clients/typescript/generated.ts
hardbyte: dropped from the `Request` DTO

reflectapi-demo/clients/typescript/generated.ts
hardbyte: Three options claude has looked into:

**A. `Client` gains `fn base_url(&self) -> &Url`** and `Request` carries `path: String`. The bare `impl Client for reqwest::Client` we ship goes away — replaced by a `ReqwestClient { inner, base_url }` wrapper (~20 lines, ours to maintain). User transports get one extra method. Generated `Interface<C>` no longer needs `base_url` at all — collapses the duplication you can see in `client-rs/src/generated.rs`.

**B. `request()` gains a `base_url: &Url` parameter.** Transports stay stateless, but the trait signature widens. Bare `impl Client for reqwest::Client` still works.

**C. Status quo.** `Request { url, headers, body }`. Codegen pre-resolves the URL on the wrapper.

My lean is **A**: we're already wrapping the transport at construction (e.g. `IntegrationsServerClient::try_new(client, base_url)`), so a `ReqwestClient { inner, base_url }` impl is a strictly smaller surface. Cross-language `Request` becomes uniform, generated code shrinks.

hardbyte: implemented in https://github.com/thepartly/reflectapi/pull/139

PR #139, opened 2026-05-08
feat(rt): move base URL onto Client transport, Request carries path

hardbyte:
Follow-up to #137 per @avkonst's `:15` thread asking what it'd take to switch the Rust `Request` DTO to `path: String` for cross-language consistency. Targets `client-request-response-shape` so the diff stays focused; will retarget to `main` once #137 merges.

## What changes

```rust
pub trait Client {
    type Error;
    fn base_url(&self) -> &Url;                         // new
    fn request(&self, request: Request) -> impl Future<...>;
}

pub struct Request {
    pub path: String,        // was: pub url: Url
    pub headers: http::HeaderMap,
    pub body: bytes::Bytes,
}
```

The bare `impl Client for reqwest::Client` / `impl Client for reqwest_middleware::ClientWithMiddleware` are gone — reqwest itself doesn't carry a base URL, so we ship thin wrappers:

```rust
pub struct ReqwestClient { pub inner: reqwest::Client, pub base_url: Url }
pub struct ReqwestMiddlewareClient { pub inner: reqwest_middleware::ClientWithMiddleware, pub base_url: Url }
```

Each `Client::request` joins `self.base_url` with `request.path` and sends.

## Codegen

- Generated `Interface<C>` no longer holds a `base_url: Url` field, and no longer threads it into every nested sub-interface. Net: ~3000 lines of generated code disappear across the snapshot suite.
- Constructor renamed `try_new(client, base_url) -> Result<Self, _>` to `new(client) -> Self`. The base URL is configured once when the transport is constructed.
- Generated method bodies just pass `&str` path: no more `self.base_url.join(path).expect(...)` per call.

## Migration

```rust
// before
let api = MyClient::try_new(reqwest::Client::new(), base_url)?;

// after
let api = MyClient::new(reflectapi::rt::ReqwestClient::new(
    reqwest::Client::new(),
    base_url,
));
```

Downstream wrappers that expose their own `try_new(client, base_url)` keep the same outer signature; only the inner construction of `Interface<C>` changes (one line).

## Why now

PR #137 made the per-language transport DTOs consistent (`Client{Request,Headers,Response}` → `{Request,Headers,Response}`, no `method`), but Rust's `Request` was still `{ url, headers, body }` while TS / Python carried `path`. Per the conversation in #137, this is the structural change that closes that gap.

Bonus: the cleanup also collapses real duplication in the generated client. Every nested sub-interface previously cloned the same `base_url: Url`, and every method did `self.base_url.join(path)?` per call. That's gone.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace` (76 demo snapshot files updated, all reflectapi unit tests green, demo binary builds)
- [ ] Confirm downstream consumers compile after migrating to `ReqwestClient` / `ReqwestMiddlewareClient`


reflectapi/src/codegen/rust.rs
```diff
@@ -1121,18 +1116,18 @@ fn __interface_types_from_function_group(
...
-            serde_name: field.0.into(),
-            description: "".into(),
-            type_: field.1.into(),
-            optional: false,
-            flatten: false,
-            public: false,
-        });
-    }
+    // The transport carries its own base URL via `Client::base_url`; we
+    // don't need to thread it through every nested interface anymore.
```
avkonst, line 1120: typical AI comment, right?

reflectapi/src/rt.rs
```diff
@@ -276,18 +253,31 @@ where
     }
 }
 
+fn __serialize_headers<H: serde::Serialize>(
+    headers: H,
+) -> Result<http::HeaderMap, (String, ProtocolErrorStage)> {
+    __serialize_headers_into(headers, http::HeaderMap::new())
+}
+
```
avkonst, line 261: I think this function is redundant, the caller can call __serialize_headers_into straight away

reflectapi/src/rt.rs
```diff
@@ -426,13 +441,35 @@ impl Client for reqwest::Client {
...
 }
 
+/// Built-in [`Client`] adapter wrapping a
+/// [`reqwest_middleware::ClientWithMiddleware`] with a base URL.
+#[cfg(feature = "reqwest-middleware")]
+#[derive(Clone, Debug)]
+pub struct ReqwestMiddlewareClient {
+    pub inner: reqwest_middleware::ClientWithMiddleware,
+    pub base_url: Url,
+}
```
avkonst, line 451: looks like this can be a single struct with templated inner of type T

reflectapi-demo/clients/rust/src/main.rs
```diff
@@ -5,7 +5,7 @@ use reflectapi_demo_client_generated::types::myapi::model::Kind;
 use reflectapi_demo_client_generated::types::myapi::proto::Headers;
 use reflectapi_demo_client_generated::DemoServerClient;
 
-type Client = DemoServerClient<reqwest::Client>;
+type Client = DemoServerClient<reflectapi::rt::ReqwestClient>;
```
avkonst, line 8: I think needs docs update for this.

reflectapi-demo/clients/rust/src/main.rs
```diff
@@ -26,11 +26,10 @@ fn pet(name: &str, kind: Kind) -> Pet {
 
 #[tokio::main]
 async fn main() {
-    let client: Client = DemoServerClient::try_new(
+    let client: Client = DemoServerClient::new(reflectapi::rt::ReqwestClient::new(
         reqwest::Client::new(),
         "http://localhost:3000".parse().unwrap(),
```
avkonst, line 31: check for base url to be not absolute path is gone?

reflectapi-demo/clients/rust/src/main.rs
```diff
@@ -26,11 +26,10 @@ fn pet(name: &str, kind: Kind) -> Pet {
 
 #[tokio::main]
 async fn main() {
-    let client: Client = DemoServerClient::try_new(
+    let client: Client = DemoServerClient::new(reflectapi::rt::ReqwestClient::new(
```
avkonst, line 29: can the generated DemoServerClient hide this adapter nuance inside, without exposing to a user? so
a client would call as before: `DemoServerClient::new(reqwest::Client::new(), "http://localhost:3000".parse().unwrap())` but the generated code would encapsulate the client and the base url into the adapter object inside of the DemoServerClient::new function? That way the adapter would become internal to the generated client.

PR #160, opened 2026-05-22
Harden Python codegen formatting

hardbyte:
## Summary
- keep Python codegen's missing-Ruff fallback to basic formatting
- fail Python codegen when `ruff format` is found but exits non-zero
- make `--format=false` a working escape hatch while preserving bare `-f/--format`
- render multi-field Rust tuple structs as Python `RootModel[tuple[...]]` instead of invalid numeric fields
- document Ruff as a generated-code formatter dependency for consistent output and pin the Rust CI job to the same Ruff version as the Python smoke test

## Validation
- `cargo fmt --check`
- `cargo test -p reflectapi-cli`
- `cargo test -p reflectapi --lib`
- `cargo run -p reflectapi-cli --quiet -- codegen -s reflectapi-demo/reflectapi.json -o /tmp/reflectapi-no-ruff-check-client -l python --python-package-name api_client --python-sync -f` with Ruff absent from PATH
- `cargo test -p reflectapi-demo tests::basic::test_reflectapi_struct_tuple -- --nocapture` with Ruff on PATH
- `cargo test -p reflectapi-demo test_python_split_modules -- --nocapture` with Ruff on PATH
- `cargo test -p reflectapi-demo --test codegen_coverage -- --nocapture` with Ruff on PATH
- `cargo test -p reflectapi-demo tests::write_python_client -- --nocapture` with Ruff on PATH

### DIFF
--- a/reflectapi-cli/tests/output_paths.rs
+++ b/reflectapi-cli/tests/output_paths.rs
@@ -30,6 +30,15 @@ fn run(args: &[&str]) -> std::process::Output {
         .expect("spawn reflectapi")
 }
 
+#[cfg(unix)]
+fn run_with_path(args: &[&str], path: &std::path::Path) -> std::process::Output {
+    Command::new(cargo_bin())
+        .args(args)
+        .env("PATH", path)
+        .output()
+        .expect("spawn reflectapi")
+}
+
 fn write_minimal_python_schema(path: &std::path::Path, type_name: &str) {
     let schema = format!(
         r#"{{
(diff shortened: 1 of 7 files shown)


reflectapi-cli/tests/output_paths.rs
```diff
@@ -30,6 +30,14 @@ fn run(args: &[&str]) -> std::process::Output {
         .expect("spawn reflectapi")
 }
 
+fn run_with_path(args: &[&str], path: &std::path::Path) -> std::process::Output {
```
avkonst, line 34: should it be only for tests cfg and namespace? as it is really only used for tests

hardbyte: I think it is already scoped just to tests being in the tests module - but will also scope it to just `cfg(unix)`

Issue #129, opened 2026-03-29
Architecture: move codegen backends toward SemanticSchema as the common backend contract


hardbyte:
Follow-up to #96.

## Summary

`#96` introduced the compiler-oriented foundations: stable symbol IDs, normalization, and `SemanticSchema`.

Today, backend consumption is still split:

- Python partially consumes `SemanticSchema` and raw `Schema`
- Rust mostly renders from raw `Schema` after backend-local preprocessing
- TypeScript mostly renders from raw `Schema` after backend-local preprocessing
- OpenAPI walks raw `Schema` directly

That means we still have multiple backend contracts and multiple places where semantic fixes can fail to propagate consistently.

This issue is for moving the architecture toward a model where all backends consume `SemanticSchema` directly, or at minimum consume a single backend-facing IR derived from it.

Recent work on the Python branch moved a little further in this direction:
- stable IDs are now assigned in the compiler path instead of constructors
- reflected Python metadata was trimmed so only irreducible Python-specific cases stay attached, while trivial cases fall back to backend inference

That is the kind of boundary this issue should make explicit across all backends.

## Why

The main benefits are architectural and compiler-facing:

- One canonical backend contract instead of a mix of raw-schema and semantic-schema paths
- Less backend drift in ordering, naming, dependency handling, and type resolution
- Fewer backend-local schema mutations and ad hoc preprocessing steps
- Stronger guarantees that symbol identity, dependency analysis, and normalization semantics are applied consistently across languages
- Better foundation for future transforms like monomorphization, richer validation, and backend-specific lowering passes
- A clearer rule that shared meaning should be inferred from schema / semantic IR first, with language-specific metadata only for irreducible cases

This is also the natural continuation of the direction described in #96: shared frontend/compiler stages with thinner, more predictable backends.

## Current pain points

- Python still needs to synchronize semantic ordering with raw schema lookups and some raw-schema mutation
- Rust and TypeScript still rely on raw schema traversal after local consolidation
- OpenAPI still bypasses the semantic layer entirely
- `Schema` phase boundaries are implicit because important transforms are performed in-place and repeated in backend code
- Compiler concerns like stable symbol identity are still stored on raw schema structs, even though they are primarily needed by normalization/codegen
- Some backends still maintain local standard-type mapping logic that partly duplicates information already present in schema shape, fallback, or normalization results

## Proposed direction

1. Define the desired common backend contract explicitly.
Either:
- all backends consume `SemanticSchema`, or
[...]

PR #168, opened 2026-06-10
feat: add #[reflectapi(hidden)] attribute for fields

avkonst:
Fields marked with `hidden` remain in the schema (with `"hidden": true`) and are functional at runtime (e.g. header extraction by the axum adapter), but are excluded from generated clients, docs, and OpenAPI specs.

This is distinct from `#[reflectapi(skip)]` which fully excludes the field from the schema (for fields whose types don't implement reflectapi traits).


reflectapi/src/codegen/python.rs
```diff
@@ -743,7 +743,11 @@ fn render_struct_with_flattened_internal_enum(
 
     // Collect the parent struct's non-flattened fields + any flattened struct fields
     let mut base_fields: Vec<templates::Field> = Vec::new();
-    for field in struct_def.fields.iter().filter(|f| !f.flattened()) {
+    for field in struct_def
+        .fields
+        .iter()
+        .filter(|f| !f.flattened() && !f.hidden())
```
hardbyte: This filter (and the others added in this file) only covers the flatten-specific render paths. The main non-flatten path `render_struct` (around line 3855) filters only on `flattened()`, so for an ordinary struct the hidden field still appears in the generated Python model. Verified on this branch:

```python
class ReflectapiDemoTestsBasicPyHiddenPlain(BaseModel):
    visible: str = Field(serialization_alias="_visible", validation_alias="_visible")
    secret: str | None = Field(  # hidden field, leaked
        default=None, serialization_alias="_secret", validation_alias="_secret"
    )
```

Also unfiltered and verified leaking:

- `collect_flattened_fields` (around line 3588): hidden fields inside a `#[serde(flatten)]` inner struct leak into the outer model
- the tuple-struct branches in `render_struct`
- all enum variant rendering (both internally and externally tagged), while the TypeScript output correctly omits the same field

Rather than chasing each site individually, see the review summary for a suggestion to strip hidden fields once at the schema level so backends cannot leak by omission.

reflectapi-demo/src/tests/basic.rs
```diff
@@ -643,3 +644,23 @@ struct TestStructWithExternalGenericTypeFallback {
...
+    struct HeadersWithHidden {
+        /// Authorization header
+        _authorization: String,
+        /// Internal tracking header, hidden from clients
+        #[reflectapi(hidden)]
+        #[serde(default)]
+        _x_internal_trace_id: String,
+    }
+
+    assert_builder_snapshot!(reflectapi::Builder::<()>::new()
```
hardbyte, line 662: `assert_builder_snapshot!` generates schema + TypeScript + Rust + OpenAPI snapshots, but not Python (unlike `assert_input_snapshot!`, `assert_output_snapshot!`, and `assert_snapshot!`, which all include Python). So this test never exercises the Python generator, which is why the Python leaks went unnoticed in CI.

Suggest adding Python generation to `assert_builder_snapshot!`, or adding an `assert_input_snapshot!` test for a struct with a hidden field. Coverage for hidden fields inside flattened structs and enum variants would be valuable too, since those are separate code paths in the Python backend.

reflectapi-derive/src/derive.rs
```diff
@@ -423,6 +432,7 @@ fn visit_field(cx: &Context, field: &ast::Field<'_>) -> Option<reflectapi_schema
         ReflectType::Output => field.attrs.skip_serializing_if().is_none(),
     };
     field_def.flattened = field.attrs.flatten();
+    field_def.hidden = attrs.hidden;
```
hardbyte, line 444: There is no guard against combining `hidden` with a required input field (one without `#[serde(default)]`). Verified on this branch: the schema then carries the contradictory pair `"required": true, "hidden": true`, generated clients and the OpenAPI spec omit the field entirely, and serde rejects exactly the payload a generated client produces (`missing field '_secret'`). Every request from a generated client fails at runtime.

Since `required` is computed a few lines above this, emitting a compile error for hidden + required on input types would catch this at the right time. The new test and the docs example both carefully pair `hidden` with `#[serde(default)]`, but nothing currently enforces it.

reflectapi/src/codegen/typescript.rs
```diff
@@ -1079,12 +1079,14 @@ fn fields_to_ts_fields(
...
             fields
                 .iter()
+                .filter(|f| !f.hidden)
                 .map(|f| field_to_ts_field(f, schema, implemented_types))
                 .collect::<Vec<_>>(),
         ),
         reflectapi_schema::Fields::Unnamed(fields) => templates::Fields::Unnamed(
             fields
                 .iter()
+                .filter(|f| !f.hidden)
```
hardbyte: Filtering hidden fields out of `Fields::Unnamed` shifts tuple positions and breaks wire compatibility. Verified on this branch with `struct TupleWithHidden(String, #[reflectapi(hidden)] u32, bool)`:

- wire format (serde): `["a",42,true]`
- generated TS type: `[string, boolean]`
- generated Rust client: `pub struct TupleWithHidden(String, bool);`

Both clients disagree with the server in both directions (requests are missing an element, responses put `true` at an index typed as something else). Since there is no coherent semantic for hiding a positional element, suggest rejecting `hidden` on unnamed fields in the derive instead of filtering here.

docs/src/getting-started/attributes.md
```diff
@@ -0,0 +1,67 @@
...
+| `#[reflectapi(output_skip)]` | Exclude the field from the output schema only. |
+| `#[reflectapi(hidden)]` | Keep the field in the schema (marked `"hidden": true`) but exclude it from generated clients, documentation, and OpenAPI specs. Useful for header fields that the server needs at runtime but clients should not see. |
+
+### `skip` vs `hidden`
+
+Both attributes remove a field from generated clients. The key difference:
+
+- **`skip`** removes the field from the schema entirely. The field's type is never reflected, so it does not need to implement `Input` or `Output`. Use this for internal bookkeeping fields whose types are not part of your API.
+
+- **`hidden`** keeps the field in the schema with a `hidden` flag. The type must still implement the relevant trait. At runtime, the axum adapter will still extract and deserialize the field (e.g. for headers). Use this for fields that are functionally required on the server side but should not appear in client interfaces or documentation.
```
hardbyte: Two caveats worth adding here:

1. For output types, `hidden` is not redaction: serde still serializes the field, so the data goes over the wire in every response (verified: the server sends `{"visible":"v","secret":"s3cret"}` while the generated TS interface only has `visible`). It is hidden from generated code and docs only. Worth stating explicitly so nobody reaches for it to hide sensitive response data.
2. A hidden input field must have `#[serde(default)]` (or be optional), otherwise generated clients cannot produce a request the server will accept. The example shows this but the text never states it as a requirement.

It would also help to spell out what retaining the field in the schema JSON actually buys over `skip`, since serde-driven runtime extraction works identically with both.

hardbyte (changes requested):
This review was carried out with Claude; every finding below was reproduced empirically against the head commit (933b7f3) rather than just read from the diff.

## What's good

- The schema-level design is clean and backwards compatible: an additive `hidden` flag that defaults to false and is omitted when serializing the schema. Threading it through `PartialEq` / `Hash` via exhaustive destructuring means the compiler will catch future omissions when fields are added.
- TypeScript, Rust client, and OpenAPI filtering are correct, including enum variants. OpenAPI header parameters get filtered for free because `convert_headers` routes through `convert_struct`.
- The attributes reference doc is a genuinely useful addition beyond this feature, and the `skip` vs `hidden` comparison is well written.
- Good drive-by fix of `test_reflectapi_struct_one_basic_field_string_reflectapi_both_equally2`, which was snapshotting the wrong type.

## Blocking

**1. Python codegen still leaks hidden fields in the common paths.** The filters added in `python.rs` only cover the flatten-specific render functions. Verified leaks on this branch (details inline on the file):

- plain structs (no flatten): `render_struct` filters only on `flattened()`, so the hidden field appears in the generated pydantic model
- flattened inner structs: `collect_flattened_fields` does not filter
- enum variants: no filtering at all (leaks for both internally and externally tagged enums)

**2. The new test cannot catch this.** `assert_builder_snapshot!` generates schema, TypeScript, Rust, and OpenAPI snapshots but not Python, so the only test for this feature never exercises the Python generator. That is why CI is green despite the above.

## Design suggestion: filter once at the schema boundary

The root cause of blocking item 1 is architectural rather than a one-off oversight. As implemented, "hidden" is an invariant that every codegen backend (and every future backend) must independently remember at every field-iteration site; `python.rs` alone has 20+ such sites. I would suggest inverting it: add a `strip_hidden()` transformation in `reflectapi-schema` (similar in spirit to `fold_transparent_types`) and apply it once at each codegen entry point. Then no backend can leak by omission, all the per-backend filters can be deleted, and any tooling that wants to see hidden fields can read the raw schema. This would resolve all three Python leaks by construction.

## Other suggestions

- **Guard `hidden` on required input fields at compile time.** Without `#[serde(default)]` the contract is unrepresentable: clients cannot send the field but the server requires it, so every generated-client request fails at runtime. Details inline on `derive.rs`.
- **Reject `hidden` on tuple/unnamed fields.** Hiding a positional element shifts indices and breaks wire compatibility in both generated clients. Details inline on `typescript.rs`.
- **Docs: `hidden` is not redaction.** For output types serde still serializes the field onto the wire; it is only hidden from generated code and docs. Worth an explicit caveat. Details inline on the docs page.
- **Clarify the motivation vs `skip` in the docs/PR description.** Runtime extraction works with `skip` too, since serde is unaffected by either attribute. The practical delta of `hidden` is that the field stays visible in the schema JSON, so it would help to state what consumes that, as it is the feature's main justification.

--claude (fable)

hardbyte:
One more architectural note: this PR connects directly to #129. "Excluded from generated clients" is shared, backend-independent meaning, and implementing it as render-time filters inside each backend is exactly the drift pattern that issue describes. The verified Python leaks are a concrete example of the failure mode: three backends filtered correctly, the one with the most raw-schema traversal sites missed several.

The `strip_hidden()` transformation suggested in the review is effectively a first increment of #129's direction: resolve visibility once in a shared stage before codegen, so no backend can leak by omission (including OpenAPI, which bypasses the semantic layer entirely today). When backends later migrate to the `SemanticSchema` contract, that strip just moves into normalization and the per-backend filters never need to exist.

--claude (fable)

reflectapi/src/codegen/python.rs
avkonst: addressed as claude suggested above

reflectapi-derive/src/derive.rs
avkonst: addressed with compile time check

reflectapi-demo/src/tests/basic.rs
avkonst: uplifting testing for python codegen to the same standard and code samples as other codegen is outside of scope of this ticket

docs/src/getting-started/attributes.md
avkonst: clarified

reflectapi/src/codegen/typescript.rs
avkonst: done

avkonst:
reverted the decision to enforce serde(default) on hidden and skipped fields - it needs to be developers decision. Documentation updated accordingly.

hardbyte (approved):
All findings addressed: schema-level strip_hidden_fields covers every backend path, tuple fields are a compile error with trybuild coverage, and the serde(default) trade-off is now clearly documented as a deliberate choice. Python sample coverage lands separately in #170.

PR #170, opened 2026-06-11
fix(codegen/python): emit nullable Option<T> fields as optional in flattened tagged-enum models

### LINKED ISSUE #167: Bug: reflectapi Python codegen emits a nullable Option<T> field as a required key
# Bug: reflectapi Python codegen emits a nullable `Option<T>` field as a *required* key

**Library:** reflectapi / reflectapi-cli — Python client code generator
**Severity:** High — generated clients raise on valid payloads.

## Summary

For a nullable field (`Option<T>`) whose key serde **omits when the
value is `None`**, the Python generator emits the field as a **required**
key (`field: T | None`, no default) instead of an **optional** one
(`field: T | None = None`). Deserializing a valid payload that omits the
key then raises — `pydantic.ValidationError: Field required` for the
pydantic backend, or `KeyError` for the attrs backend.

The output is also **inconsistent**: the same `Option<T>` field generates
correctly (with a `None` default) on a standalone struct, but
incorrectly (required, no default) on a struct used as a variant of an
internally-tagged enum.

## Minimal reproduction

Two structurally-identical nullable fields — one on a standalone struct,
one on a tagged-enum variant:

```rust
use reflectapi::Output;

#[derive(serde::Serialize, serde::Deserialize, Output)]
pub struct Standalone {
    pub amount: Option<String>,
}

#[derive(serde::Serialize, serde::Deserialize, Output)]
#[serde(tag = "type", rename_all = "snake_case")]
pub enum Thing {
    Variant(VariantBody),
}

#[derive(serde::Serialize, serde::Deserialize, Output)]
pub struct VariantBody {
    pub amount: Option<String>,
}
```

Generated Python (observed):

```python
class Standalone(BaseModel):
    amount: str | None = None        # ✅ optional — correct

class VariantBody(BaseModel):
    amount: str | None               # ❌ required key, no default — BUG
```

Same divergence in the attrs backend: `Standalone.from_dict` pops with a
default (`d.pop("amount", None)`), while `VariantBody.from_dict` pops
without one (`d.pop("amount")`), so an absent key raises `KeyError`.

## Why this breaks at runtime

serde serializes `Option::None` by dropping the key (the producer emits
no `amount` field). A correct client must therefore treat the key as
optional. The required form raises on a payload that is valid by the
schema:

```python
# pydantic backend
VariantBody.model_validate({"type": "variant"})
# Before: pydantic.ValidationError — amount: Field required [type=missing]
# After:  OK, amount = None

# attrs backend
VariantBody.from_dict({"type": "variant"})
# Before: KeyError: 'amount'
# After:  OK, amount = None
```

## Expected

A nullable `Option<T>` field should generate as an optional Python key
(`T | None = None`, and `d.pop(key, None)` in the attrs `from_dict`)
**regardless of whether the owning struct is standalone or a tagged-enum
variant** — matching the standalone behavior the generator already
produces.

## Workaround

Patch the affected generated models to default the field to `None`
(pydantic: `field: T | None = None`; attrs `from_dict`:
`d.pop(key, None)`). This is clobbered on the next codegen run, hence
this report — the fix belongs in the generator so regenerated output
stays correct.

hardbyte:
Fixes #167

## Root cause

The bug is one step removed from the issue's minimal repro: a standalone internally-tagged enum already generates correctly, because that render path routes field optionality through `resolve_field_optionality` (which treats any `Option<T>` field as an optional Python key, since serde omits the key for `None`).

The failing shape is a **struct with a `#[serde(flatten)]`-ed internally-tagged enum** — e.g. `OffersSearchOffer` flattening an offer-details enum, producing per-variant models like `OffersSearchOfferProvisional`. Those models are generated by `render_struct_with_flattened_internal_enum` / `render_struct_with_flatten_standard`, which hand-rolled optionality from `field.required` alone and ignored the `Option<T>` special case. Two symptoms:

- `Option<T>` without `skip_serializing_if` → `amount: str | None` **required, no default** → `pydantic.ValidationError: Field required` on a valid payload that omits the key
- `Option<T>` with `skip_serializing_if` → doubled union `note: str | None | None = None`

@MattGson note: your minimal repro only fails when the enum is flattened into a parent struct — worth sanity-checking that matches the offers code shape.

## Fix

Route all five inline field-emission sites in `reflectapi/src/codegen/python.rs` through `resolve_field_optionality`:

- parent-struct base fields and flattened-sibling-enum fields in `render_struct_with_flattened_internal_enum`
- tuple-variant inner-struct fields in the same function
- regular fields in `render_struct_with_flatten_standard`
- flattened-enum-as-field emission (had its own double-`| None` bug for `Option<Enum>`)

Field optionality is now decided in exactly one place, so flattened-enum variant models match standalone struct output.

## Tests

- New snapshot samples for both shapes: the issue's standalone repro (guards the already-correct path) and the flattened variant that reproduces the bug. Both run through `assert_snapshot!`, covering schema JSON, TypeScript, Rust, OpenAPI, and Python.
- Per @avkonst's suggestion, python codegen is now covered by the same test samples as the other languages: it was already in `assert_input_snapshot!`/`assert_output_snapshot!`/`assert_snapshot!`; the one gap was `assert_builder_snapshot!` (the 4 namespace samples), which now snapshots python too (appended last so existing snapshot indices don't churn).
- 5 existing `generic_flatten_*` python snapshots updated — each is the same one-line correction (`if_else: X | None` gaining `= None`); those committed snapshots had been silently capturing the bug.
- Runtime-validated the generated client under pydantic with the local `reflectapi-python-runtime`: `model_validate({"type": "variant", ...})` with omitted optional keys now succeeds on variant models directly and through the discriminated-union root, and populated keys still parse.
- Full demo suite passes (193/193); fmt/clippy/build clean.


reflectapi-demo/src/tests/snapshots/reflectapi_demo__tests__enums__internally_tagged_enum_with_optional_fields-2.snap
avkonst, line 1: why so many TS and Rust snapshots are affected? I confident it is all expected?

hardbyte: yup most from new repro files. all good

Issue #169, opened 2026-06-10
Generated Rust client leaks secrets via #[instrument] — add secrecy::SecretString support


avkonst:
Root Cause:

The Rust client codegen emits #[instrument] on endpoint functions. Request structs containing password fields are logged by tracing as function arguments. Since codegen maps all opaque string-like types to plain String, the Debug impl prints passwords in cleartext to logs.

On the server side, secret::SecretString correctly redacts on Debug. But the generated client has no awareness of this — the field is just String, and #[instrument] happily prints it.

Solution:

Add first-class support for secrecy::SecretString in reflectapi, following the same pattern as uuid::Uuid and chrono::DateTime:

reflectapi crate: Add secrecy feature flag. Implement reflectapi::Input + reflectapi::Output for secrecy::SecretString. Registers as primitive "secrecy::SecretString".

Rust codegen: Recognize "secrecy::SecretString" → emit secrecy::SecretString in generated structs. Add secrecy dep to generated Cargo.toml. Since secrecy::SecretString implements Debug as SecretString([REDACTED]), #[instrument] no longer leaks values.

OpenAPI codegen: Map to { type: "string", format: "password" } (standard OpenAPI convention — tools like Swagger UI mask the input).

Python codegen: Map to pydantic.SecretStr (redacts on repr()/str()).

TypeScript codegen: Map to string (no runtime debug concern in TS; optionally a branded SecretString type for documentation).

Migration for private codebases:

Private secret::SecretString impls should set their reflectapi type to fall back to the public type. In the private Input impl for secret::SecretString, register as "secrecy::SecretString" (or use #[reflectapi(type = "secrecy::SecretString")]). This ensures the schema carries the public type name, and all codegen backends emit the correct secret-safe type.

Related: secret::SecretString is not capable of reflectapi::Output and a bunch of api responses *.signin return tokens as bare strings. These would be better converted to secret strings as well

avkonst:
I deleted the previous comment. bad idea. ignore

avkonst:
@fernando-fiddymont 

More of my thoughts on this:

* I think it is right that a secret type does not implement Serialize, it would force a developer to use the attribute serde(serialize_with = Secret::dangerous_serialize) - we can require it on the server side and generate it on the client side
[...]

PR #173, opened 2026-06-11
fix(codegen/rust): stop recording request bodies in instrumented client spans

hardbyte:
Part of #169  (just the mitigation); first-class `secrecy::SecretString` support (type mapping across all backends) remains as a follow-up.

## Problem

Instrumented Rust clients emit:

```rust
#[tracing::instrument(name = "/auth.login", skip(self, headers))]
pub async fn login(&self, input: LoginRequest, headers: Empty) -> ...
```

`input` is not skipped, so tracing records the entire request body as a span field via its `Debug` impl. Since codegen maps string-like types to plain `String`, credentials print to logs in cleartext.

## Fix

The generated attribute now uses `skip_all`. No function arguments are recorded; the span keeps the endpoint path as its name. This protects **every** request field — including secrets typed as plain `String` on the server — which first-class `SecretString` support alone would not.

Trade-off: request bodies no longer appear as span fields at all. Given the payload is arbitrary user data with no redaction story, recording it by default was a liability rather than a feature.

## Tests & validation

- New test `instrumented_rust_client_does_not_record_request_bodies` generates the demo schema with `.instrument(true)` and asserts every emitted `#[tracing::instrument]` attribute uses `skip_all` (and that attributes are emitted at all).
- The committed demo Rust client is regenerated (8 endpoints switch to `skip_all`) and compiles as a workspace member.
- Empirically validated with a scratch crate replicating the exact generated method shape under a captured `tracing-subscriber`:
  - old attribute: `INFO /auth.login{input=LoginRequest { username: "brian", password: "hunter2-s3cret" }}: sending request` — password in cleartext
  - new attribute: `INFO /auth.login: sending request` — nothing recorded, span name retained
- Full demo suite passes (195/195); fmt/clippy/build clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- **Security:** instrumented Rust clients (`--instrument`) no longer record request bodies and headers as tracing span fields. The generated `#[tracing::instrument]` attribute now uses `skip_all`, so credentials in request payloads (passwords, tokens) can no longer reach logs in cleartext via their `Debug` output. Spans are still named after the endpoint path. Regenerate clients to pick this up.
+
 ## 0.17.6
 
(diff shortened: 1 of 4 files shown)


avkonst:
I thought about doing this but then concluded the better more complete fix would be to tell via reflectapi attribute per struct or per field what to include into the instrumentation. But that is bigger change so I parked if for now.
Do you want to merge it to close the leak of credentials in traces?

PR #175, opened 2026-06-16
feat(codegen/rust): add deprecated attribute for enum variants and types

akozlev:


hardbyte:
1. reflectapi/src/codegen/rust.rs:398 and reflectapi/src/codegen/rust.rs:857: generated clients can fail under -D warnings after adding #[deprecated] to generated types. The generated interface methods still reference those deprecated types in their own signatures, so compiling the generated client itself emits deprecated warnings before any downstream caller has used the deprecated API. A crate with #![deny(warnings)] will fail just by compiling the generated code. The generated module likely needs internal #[allow(deprecated)] around generated references while still exposing #[deprecated] on the public items so downstream users get warnings when they use them.
2. reflectapi/src/codegen/rust.rs:402, reflectapi/src/codegen/rust.rs:620, reflectapi/src/codegen/rust.rs:732: deprecation notes are interpolated directly into Rust string literals. Notes containing quotes, backslashes, or newlines generate invalid Rust. Example source #[deprecated(note = "use \"New\"")] becomes #[deprecated(note = "use "New"")]. This should use a Rust string literal escaping helper rather than raw interpolation.
3. reflectapi/src/codegen/schema/semantic.rs:82 and reflectapi/src/codegen/schema/semantic.rs:145: type/variant deprecation metadata is added to raw schema but not to SemanticStruct, SemanticEnum, or SemanticVariant. In the context of #129, this puts new backend-relevant meaning outside the semantic contract and guarantees drift as backends move toward SemanticSchema. The current snapshots already show the drift: Rust renders deprecated type/variant attributes, but TypeScript/Python/OpenAPI snapshots for the same deprecated types/variants do not surface the metadata.

avkonst:
regarding 1, is not the point of deprecation to issue a warning and let a user to decide if they want to fail on this, leave a warning or fix? I think they can steer the warning to error or nothing on their side?

avkonst (approved):

hardbyte:
Following up on my three earlier points now the fixups are in:

1. **`-D warnings`** — the `#[allow(deprecated)]` on the generated module addresses it. Andrey, agreed it's the consumer's call how to treat deprecation warnings; the problem was only that the generated crate would warn on its *own* internal references before any consumer touched the deprecated item. Resolved.
2. **Raw interpolation of the note** — still present (`format!("#[deprecated(note = \"{note}\")]")`). A note containing `"` or `\` produces code that doesn't compile. `{note:?}` is the one-token fix: `Debug` for `str` emits a quoted, escaped Rust string literal, so `#[deprecated(note = {note:?})]` is always valid. I'd like this in before merge.
3. **Semantic layer** — fine as a follow-up, but the same sample's TS / Python / OpenAPI snapshots don't surface the metadata, so a tracking issue would be good.

The docs-preview failure on this branch looks unrelated to the change; a re-run should tell.

— Fable on behalf of Brian

avkonst (approved):

PR #176, opened 2026-06-19
advance transform attribute to access full field metadata

avkonst:


docs/src/getting-started/attributes.md
```diff
@@ -22,9 +22,39 @@ ReflectAPI provides `#[reflectapi(...)]` attributes that control how Rust types
...
+
+A field may use an Option wrapper at the Rust level (e.g., because a middleware populates it before your handler runs) but you want generated clients to see the unwrapped, required type.
+
+```rust,ignore
+#[derive(serde::Deserialize, reflectapi::Input)]
+struct MyRequest {
+    /// In Rust this is `Option<String>` (populated by middleware),
+    /// but clients see it as a required `String` field.
+    #[serde(default)]
+    #[reflectapi(input_transform = "reflectapi::transforms::make_required")]
```
JosiahBull, line 55: nice

docs/src/getting-started/attributes.md
```diff
@@ -22,9 +22,39 @@ ReflectAPI provides `#[reflectapi(...)]` attributes that control how Rust types
...
+    pub tenant_id: Option<String>,
+}
+
+#[derive(serde::Deserialize, reflectapi::Input)]
+struct PatchRequest {
+    /// In Rust this is `reflectapi::Option<String>` (populated by middleware),
+    /// but clients see it as a required `Option<String>` (nullable, not omittable).
```
JosiahBull, line 62: Subtle difference based on `reflectapi::Option` vs `std::Option`. As a developer I'd typically expect `make_required` to do the same regardless of what Option variant it's on - would be a bit of a footgun imo.

JosiahBull (changes requested):

docs/src/getting-started/attributes.md
avkonst: making std option required means removing Undefined and None variants for the non-nullable field.

making reflectapi required means removing Undefined variant, as None is expected possible null value.

the difference is on purpose as the semantics of both types required.

however, the client can provide their own transform functions.

avkonst:
@JosiahBull what change are you requesting to make?

docs/src/getting-started/attributes.md
JosiahBull: If I use a function called `make_required` on something named `Option` I would expect the generated client to disallow all None-alike values (None, null, etc).

JosiahBull: I think we need to rename something here to make this clearer

avkonst: Do you know that from reflectapi (and our API standard) required is not equal to not-nullable? These are two distinct things. It is possible to have non-required but nullable and required but nullable and 2 other combos. So the transform function does exactly that: it removes the declaration for the field to be required, which is for non nullable field means no None of the standard Option and for nullable field it means no Undefined of the reflectapi Option.

avkonst: we can add make_nonnullable, and super set of both make_required_and_nonnullable

JosiahBull: I would be happy with this outcome, will mark as approved.

JosiahBull (approved):

PR #179, opened 2026-09-10
feat(derive): #[reflectapi(header)] sends a response field as a header

elilabes:
A handler returns a body and a status and nothing else, so a route cannot set a `Set-Cookie` — which makes session establishment the one thing a reflectapi API cannot serve, and it is usually the first route a browser calls.

A response field marked `#[reflectapi(header)]` is now lifted onto the response instead. The header name is the field name in kebab-case, since a Rust identifier cannot hold a `-`; a serde rename is used verbatim.

Header fields stay out of the schema, so generated clients are identical whether or not a route sets one. That is the design rather than a shortcut: a browser filters `Set-Cookie` out of what a client can read, so describing it would generate a field that is always empty — and a field carried in the body type's field list would mislead anything reading the schema from outside this repo. Describing headers properly is a separate feature and wants its own type reference beside `input_headers`, not a flag on body fields.

Two guards, because both failures are silent otherwise. `#[serde(skip_serializing)]` is required — the derive cannot make serde skip a field, and without it the value goes in the body as well as the header, which for a sealed session would defeat the point. And the field type must implement `HeaderValue`, provided for `String`, `Option<T>` (sends nothing) and `Vec<T>` (sends one header per element, which `set-cookie` routinely needs).

Error types carry header fields too, so a refusal can clear a cookie.


docs/src/getting-started/quick-start.md
```diff
@@ -111,6 +111,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 ```
 
+## Setting Response Headers
```
avkonst, line 114: this does not belong here. It needs to go to the attributes.md doc

reflectapi-demo/tests/errors/header_without_skip_serializing.rs
```diff
@@ -0,0 +1,10 @@
+#[derive(serde::Serialize, reflectapi::Output)]
+struct Response {
+    user_id: String,
+    // Without `#[serde(skip_serializing)]` this value would be sent in the
```
avkonst, line 4: do they need to be mutually exclusive?
I think it is best to leave it to a user to decide, rather than enforce?

reflectapi-derive/src/derive.rs
```diff
@@ -484,3 +491,109 @@ fn visit_name<'a>(
     }
     (result.into(), String::new())
 }
+
+/// The `reflectapi_response_headers` override for a struct with
+/// `#[reflectapi(header)]` fields, or nothing when it has none.
+///
+/// The field is left out of the schema by the same rule that leaves out any
```
avkonst, line 498: this is interesting question... 
I guess you picked the option to not expose headers in the schema, docs, and generated clients. Maybe it is the way to go. Decision to make but would be good to explore pros and cons.

reflectapi/src/builder/handler.rs
avkonst, line 460: does this need to be updated as well? plain handler_wrap is updated

reflectapi/src/builder/handler.rs
```diff
@@ -376,6 +376,37 @@ where
...
+        };
+        for (name, value) in declared_headers {
+            match (
+                http::HeaderName::from_bytes(name.as_bytes()),
+                http::HeaderValue::from_str(&value),
+            ) {
+                (Ok(name), Ok(value)) => {
+                    response_headers.append(name, value);
+                }
+                _ => {
```
avkonst, line 394: Prefer explicit `Err(name_err), Err(value_err)`. And maybe make use of it in the throw case

reflectapi/src/traits.rs
```diff
@@ -53,6 +53,54 @@ pub trait Input {
...
     fn reflectapi_output_type(schema: &mut crate::Typespace) -> crate::TypeReference;
+
+    /// The response headers this value carries, as `(name, value)` pairs.
+    ///
+    /// Overridden by the derive for types with `#[reflectapi(header)]` fields;
+    /// every other type answers with none. A name repeated across pairs is a
+    /// header sent more than once, which `set-cookie` routinely is.
+    fn reflectapi_response_headers(&self) -> Vec<(&'static str, String)> {
+        Vec::new()
+    }
```
avkonst, line 64: presence of this code (and all of of the impls below), makes me preferring the `header` attribute to get first class support in the schema. So reflectapi_output_type would be the only we needed as the field definition in the schema would have it. Possible?

reflectapi-derive/src/derive.rs
```diff
@@ -484,3 +491,109 @@ fn visit_name<'a>(
...
+        .rename_all_rules()
+        .serialize
+        .apply_to_field(&base);
+
+    if serialized != rule_applied {
+        serialized.to_owned()
+    } else {
+        RenameRule::KebabCase.apply_to_field(&base)
+    }
+}
```
avkonst, line 599: same comment. I think better to carry header attribute for a field to the schema using the engine which is already there. Then this code should not be needed.

avkonst (changes requested):

avkonst:
Thank you Eli. You picked (not sure deliberately or not) a side path for the attribute get aside of the schema. And I think it adds parallel flow and breaks the consistency of the internal architecture. I think the attribute should make its way to the schema, then the rest of the ecosystem would make use of it (i.e. you start with the handler_warp first).
It is still fine to restrict the new attribute to be used with skip_serializing, if you want to but it does not have to be this way.
Your implementation assumes that attribute can be used on a certain set of types (which I believe traits are for), but it is better to reuse the existing ecosystem of headers handling from the request. As far as I remember these work with any types, without need for new traits.

hardbyte (commented):
I've read this alongside #180, since both are about headers the schema doesn't describe. I agree with Andrey's direction, and want to add a concrete alternative shape, because I think the constraints in this PR (`skip_serializing` required, a `HeaderValue` trait, a header-naming rule that ignores `rename_all`) are all symptoms of putting headers *inside* the body type.

The input side already has the answer: headers are a separate type on the handler (`input_headers: TypeReference` on `Function`), so they get their own place in the schema and every generator renders them as a headers struct. Mirror that for output: `output_headers: Option<TypeReference>` on `Function`, plus a way for the handler to return them, e.g.

```rust
async fn sign_in(state, req: SignInRequest, _: reflectapi::Empty)
    -> Result<reflectapi::WithHeaders<SignedIn, SessionHeaders>, SignInError>
```

where `SessionHeaders: reflectapi::Output` is an ordinary struct (`set_cookie: Vec<String>`, `x_request_id: Option<String>`). What that buys:

- No `skip_serializing` coupling: the body type is just the body type, and the derive doesn't have to inspect serde attributes to keep a field out of the body.
- No new `HeaderValue` trait: the same header-struct machinery that serialises `input_headers` on the client side handles arbitrary field types via serde. `Vec<T>` → repeated header is the one rule to add.
- Header names: the struct's own serde names apply (`rename` / `rename_all`), exactly as for input headers today. No special "container rename doesn't reach a header" rule to document.
- The schema knows about it, so clients can expose response headers in a typed way (#88), OpenAPI can emit `responses.*.headers`, and a browser client can simply *not surface* `set-cookie` — that's a generator decision, not a schema decision.
- Error arm: `Result<WithHeaders<T, H>, WithHeaders<E, H>>` is verbose; a `StatusCode`-style trait on the error type (`fn response_headers(&self) -> H`) would be the ergonomic version and could land as a follow-up.

On "browsers can't read `Set-Cookie`, so don't describe it": true for that one header in a browser, not for Rust or Python clients, and not for the other headers you'd reach for next (`x-request-id`, rate-limit headers, `Location`). The schema is the contract for every consumer, so the browser case belongs in the TypeScript generator, not in the contract.

Smaller notes on the diff as it stands:

- Header *names* are `&'static str` known at derive time, so `HeaderName` validity can be checked in the derive rather than turning into a 500 at request time. Values still need the runtime check.
- The second handler-wrap path Andrey pointed at needs the same treatment.
- Agree the docs belong in `attributes.md` rather than the quick start.

— Fable on behalf of Brian

PR #180, opened 2026-09-13
feat(codegen): add --required-headers for headers clients must always…

JackFawthorpePartly:
… send

Some deployments need every request to carry a header the service itself never declares — an API key checked by a gateway, a tenant id read by a routing layer. Because no handler takes them they are absent from the schema, and nothing stops a caller from forgetting them.

`--required-headers` names those headers at generation time. Generated clients cannot be constructed without them and send them on every request, as middleware around the transport, so they survive a caller-supplied transport. A header set for an individual call still wins.

TypeScript takes a `RequiredHeaders` object in `client(base, ...)`; Rust takes a `RequiredHeaders` struct in `Interface::new`, wrapping the transport in the new `reflectapi::rt::WithRequiredHeaders`; Python takes one keyword-only argument per header and installs the new `Sync`/`AsyncRequiredHeadersMiddleware`; OpenAPI emits a required `in: header` parameter per operation, leaving a header a handler already declares as the handler declared it.

BREAKING: `codegen::openapi::generate_spec` now returns `anyhow::Result<Spec>` so an invalid required header name is reported rather than silently dropped. `Spec::from(&schema)` is unchanged.


hardbyte (commented):
Thanks Jack — the problem is real (gateway API keys, tenant routing) and the implementation is careful: case-insensitive matching, per-call wins, identifier-clash detection, the Python middleware doesn't mutate the incoming request, and the tests cover the precedence rules. My concerns are about *where the contract lives* and *what shape the generated API takes*. Both are hard to change once consumers depend on them, so I'd like to settle them before this merges.

### 1. The contract lives in a codegen flag, not the schema

Everything else a client is told about the API comes from `reflectapi.json`. `include_tags` / `exclude_tags` *filter* the schema; `required_headers` *adds* contract the schema doesn't know about. Consequences:

- Two teams generating from the same schema get clients with different constructor signatures depending on their CI flags, and a schema diff never shows a header being added or removed.
- Each backend has to be taught about it separately (this PR touches all four renderers), which is the drift pattern #129 is trying to move away from — and the same point Andrey made on #179.
- OpenAPI dedupes against a handler-declared header but the clients don't: with the demo's `Headers { authorization }` and `--required-headers authorization`, TS/Rust/Python require it both at construction *and* on every call.

Suggested shape: put it in the schema. Either a schema-level list (`Builder::required_header("x-api-key")`) or, more generally, a schema-level `input_headers: Option<TypeReference>` next to the per-function one — a headers struct that applies to every endpoint. Generators then render that as a constructor-time parameter with per-call override. That also unlocks the thing people ask for most: setting `authorization` once on the client instead of on every call. If the gateway header genuinely isn't known to the service author, keep the CLI flag, but have it *transform the schema* before rendering (there's precedent now in `reflectapi-schema/src/transforms.rs`), so the renderers read one place.

### 2. Ship the runtime pieces as general-purpose default-header decorators regardless

`WithRequiredHeaders<C>` is a "default headers" transport decorator; nothing about it is specific to required-ness. Rename it (`WithDefaultHeaders`?), and export the TS equivalent from `lib.ts` as a public helper rather than `__with_required_headers`. Then a caller can do `Interface::new(WithDefaultHeaders::new(client, headers))` or `client(withDefaultHeaders(base, {...}))` today with zero codegen changes, and that solves most of the motivating problem on its own. The Python middleware is the strongest piece here: unlike the existing `headers=` kwarg it survives a caller-supplied `client=`, which deserves a line in its docstring.

Worth stating explicitly in the docs: Rust already has `reqwest::ClientBuilder::default_headers`, Python already has `headers=`. The new value of this PR is *enforcement*, not the ability to send headers, and the design should be judged on that.

### 3. Rust: the generated `Interface` type changes with the flag

`Interface::new` returning `Interface<WithRequiredHeaders<C>>` means anyone storing the client in a struct now writes `Interface<WithRequiredHeaders<ReqwestClient<reqwest::Client>>>`, and flipping the flag is a type-level break across every consumer. If enforcement stays in the constructor, consider keeping `Interface<C>` and holding the headers on the interface, or leave the wrapping to the caller via the decorator in (2).

### 4. Rust: `RequiredHeaders::new` is positional in alphabetical order of header name

Two `HeaderValue`s in an order determined by a sort the caller can't see: swapping `x-api-key` and `x-tenant-id` compiles and fails at the gateway. Struct-literal construction already exists; I'd drop `new`, or take `impl TryInto<HeaderValue>` so callers don't need the `HeaderValue::from_str(..)?` ceremony either.

### 5. TS: a second positional argument

`client(base, required_headers)` closes the door on further constructor options (timeout, retry, fetch impl). `client(base, { headers })` stays extensible; the generated `RequiredHeaders` type becomes the type of `headers`.

### 6. CI: the failing test typechecks the whole demo schema through the Rust dependency stubs

`rust_client_takes_required_headers_at_construction` fails with `cannot find __stream_request_impl in reflectapi::rt` and `cannot find type Option in crate reflectapi`. The stubs don't cover streaming endpoints or `reflectapi::Option`, and no existing test typechecks the full `crate::builder()` schema, so this test is the first to hit that. Either typecheck a small purpose-built schema like the other tests do, or extend the stubs (#174 adds `reflectapi::Option`; `__stream_request_impl` would need adding too).

### 7. Smaller things

- `generate_spec` becoming fallible is fine by me. If we'd rather keep it infallible, validating names in the CLI before calling, or a `Spec::try_from`, would do.
- Python: a required header named `timeout`, `headers`, `middleware`, `auth` or `client` generates a keyword argument that silently shadows the base-class kwarg of the same name. Worth rejecting in `required_headers::resolve`.
- Docs: the precedence rule (per-call header wins) is stated once at the top; repeat it in the Rust and Python rows of the table since that's where people will look.

**Bottom line:** I'd merge the decorators (Rust / TS / Python) as standalone runtime additions now, and reshape enforcement to flow through the schema before adding the constructor changes. Happy to pair on the schema-side design.

— Fable on behalf of Brian

avkonst:
@JackFawthorpePartly 

> Some deployments need every request to carry a header the service itself never declares — an API key checked by a gateway, a tenant id read by a routing layer. Because no handler takes them they are absent from the schema, and nothing stops a caller from forgetting them.

So, a service expects to function correctly only if there is a gateway upfront checking headers. What is the disadvantage to have these headers declared by the service in the schema? The app does not need to check it provided the gateway does (if that is the contract) but the service is aware and declares these headers as it would not function otherwise.

avkonst:
The little automation for this is to have bunk declaration at the builder level, but I am not sure what the benefits it provides.

